PR #6, opened 2023-08-02
OTEL Tracing for All Routes

sordina:
This PR adopts the patterns used by https://github.com/hasura/mongo-agent-rs for OTEL tracing.

Heavy lifting is performed by the https://docs.rs/tracing crate.

### Open Questions

Should we get these errors if OTLP endpoint not set?

```
OpenTelemetry trace error occurred. request failed with status 404 Not Found
```

### Changes

TODO:

* [x] Trace all routes
* [x] Include V3 tags - Can be done through ENV: `OTEL_SERVICE_NAME ` and `OTEL_RESOURCE_ATTRIBUTES`
* [x] Test automatic tracing
* [x] Test correct nesting of manual traces (including across async boundaries)
* [ ] Automatic tests
* [x] Documentation - See Rob's V3 docs repo, Tracing.md
* [ ] Determine if GRPC, vs. HTTP should be selected or configurable
* [ ] Talk to Shahid about MT deployment changes for ENV changes

### Configuration

When authoring an agent using this SDK, the following run-time environment variables can be used to configure OTEL tracing:

(see https://github.com/TommyCpp/opentelemetry-rust/blob/7c5ad125cec85240909243092202f200ea843960/opentelemetry/src/sdk/resource/env.rs)

* `OTEL_RESOURCE_ATTRIBUTES` e.g. `key=value, k = v , a= x, a=z`
* `OTEL_SERVICE_NAME` e.g. `ndc_hub_example`
* `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` e.g. `http://localhost:4317`

### Manual Testing

Run Jager:

```bash
docker run --name jaeger -e COLLECTOR_OTLP_ENABLED=true -p 16686:16686 -p 4317:4317 -p 4318:4318 jaegertracing/all-in-one:1.45
```

Run the example agent:

```bash
OTEL_SERVICE_NAME=ndc_hub_example OTEL_EXPORTER_OTLP_TRACES_ENDPOINT=http://localhost:4317 cargo run --bin 
ndc_hub_example -- serve --configuration <(echo 'null')
```

Request the schema:

```bash
curl localhost:8100/schema | jq .
```

See results in Jager (http://localhost:16686/): 

image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,6 @@ path = "bin/main.rs"
 axum-tracing-opentelemetry = { version = "0.11.0", features = [
     "otlp",
     "tracing_subscriber_ext",
-    "zipkin",                 # provides b3 and b3multi propagators
 ] }
 opentelemetry = "0.19.0" # must be same version axum-tracing-opentelemetry uses
 opentelemetry-otlp = { version = "0.12.0", features = [
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,9 +26,6 @@ axum = "0.6.18"
 axum-macros = "0.3.7"
 clap = { version = "4.3.9", features = ["derive", "env"] }
 ndc-client = { git = "http://github.com/hasura/ndc-spec.git", rev = "f3a6759" }
-# opentelemetry = { version = "0.18.0", features = ["rt-tokio"] }
-# opentelemetry_sdk = "0.18.0"
-# opentelemetry-otlp = "0.11.0"
 opentelemetry-semantic-conventions = "0.10.0"
 prometheus = "0.13.3"
 serde = "1.0.164"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,14 +12,23 @@ name = "ndc_hub_example"
 path = "bin/main.rs"
 
 [dependencies]
+axum-tracing-opentelemetry = { version = "0.11.0", features = [
+    "otlp",
+    "tracing_subscriber_ext",
+    "zipkin",                 # provides b3 and b3multi propagators
+] }
+opentelemetry = "0.19.0" # must be same version axum-tracing-opentelemetry uses
+opentelemetry-otlp = { version = "0.12.0", features = [
+    "reqwest-client",
+] } # must be same version axum-tracing-opentelemetry uses
 async-trait = "0.1.68"
 axum = "0.6.18"
 axum-macros = "0.3.7"
 clap = { version = "4.3.9", features = ["derive", "env"] }
 ndc-client = { git = "http://github.com/hasura/ndc-spec.git", rev = "f3a6759" }
-opentelemetry = { version = "0.18.0", features = ["rt-tokio"] }
-opentelemetry_sdk = "0.18.0"
-opentelemetry-otlp = "0.11.0"
+# opentelemetry = { version = "0.18.0", features = ["rt-tokio"] }
+# opentelemetry_sdk = "0.18.0"
+# opentelemetry-otlp = "0.11.0"
 opentelemetry-semantic-conventions = "0.10.0"
 prometheus = "0.13.3"
 serde = "1.0.164"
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -12,14 +12,20 @@ name = "ndc_hub_example"
 path = "bin/main.rs"
 
 [dependencies]
+axum-tracing-opentelemetry = { version = "0.11.0", features = [
+    "otlp",
+    "tracing_subscriber_ext",
+    "zipkin",                 # provides b3 and b3multi propagators
```
paf31, line 18: We're assuming W3C trace context propagation in V3.

src/default_main.rs
```diff
@@ -104,36 +106,7 @@ where
...
-
-    tracing_subscriber::fmt()
-        .with_max_level(tracing::Level::DEBUG)
-        .init();
-
-    let _tracer = opentelemetry_otlp::new_pipeline()
-        .tracing()
-        .with_exporter(
-            opentelemetry_otlp::new_exporter().tonic().with_endpoint(
-                serve_command
```
paf31, line 117: Is `otlp_endpoint` not used any more?

paf31: If we're changing the CLI then we'll just need to run it by the infra team, but also remove the unused args and update the Dockerfile as needed.

sordina: Don't we push rather than pull now?

Cargo.toml
sordina: 👍

src/default_main.rs
```diff
@@ -104,34 +104,12 @@ where
...
-                ),
-            ]),
-        ))
-        .install_batch(opentelemetry::runtime::Tokio);
+    // Set endpoint ENV picked up by macros in `traces` crate via CLI option if used
+    serve_command.otlp_endpoint.map(|e| {
+        env::set_var(opentelemetry_otlp::OTEL_EXPORTER_OTLP_TRACES_ENDPOINT, e);
+    });
+
+    axum_tracing_opentelemetry::tracing_subscriber_ext::init_subscribers().expect("Couldn't init OT");
```
paf31, line 113: The docs for this function say "very opinionated, read the source to implement your own" and the source is here: https://github.com/davidB/tracing-opentelemetry-instrumentation-sdk/blob/main/init-tracing-opentelemetry/src/tracing_subscriber_ext.rs

Rather than modify the environment variables to make this work, what about taking that approach and making a custom setup function?

paf31 (commented):
There are a fair few todos and commented lines here, so let me know when this is ready for a final review once you've had a chance to discuss the API with @shahidhk and I will do a final review.

sordina:
> There are a fair few todos and commented lines here, so let me know when this is ready for a final review once you've had a chance to discuss the API with @shahidhk and I will do a final review.

Will do!

src/default_main.rs
sordina: Sounds like a great idea, although I'm not sure how to do this. Mind pairing for a minute to run through what this would involve?

src/default_main.rs
```diff
@@ -198,6 +179,8 @@ where
         .route("/mutation", post(post_mutation::<C>))
         .with_state(state)
         .layer(TraceLayer::new_for_http())
+        .layer(response_with_trace_layer())
+        .layer(opentelemetry_tracing_layer())
```
paf31, line 182: I don't think we should use `opentelemetry_tracing_layer()` and `TraceLayer::new_for_http()` at the same time?

paf31:
I'm not sure about the two tracing layers. I was planning to merge this one and leave the other as a follow-up, but I'll merge the other for now, since I've tested it most recently and I'm sure there are no dropped spans. We can revisit later, but since I need to merge _something_ for tracing.

codedmart:
Closing in favor of #12

PR #8, opened 2023-08-08
Add `fetch_metrics` call for fetching metrics

danieljharvey:
When implementing metrics for Postgres NDC I found querying the connection pool a little trickier than I'd like. Because the connector had no idea when `/metrics` were being requested, I had to set up a timer thread to poll the connection pool every second or so, possibly needlessly.

This is a very rough sketch of an solution - the connector provides a function that takes the `State` plus the Prometheus `Registry` and updates all counters.

### DIFF
--- a/src/connector.rs
+++ b/src/connector.rs
@@ -180,6 +180,16 @@ pub trait Connector {
         metrics: &mut prometheus::Registry,
     ) -> Result<Self::State, InitializationError>;
 
+    /// Update any metrics from the state
+    ///
+    /// Useful to query a connection pool
+    /// and use it to update metrics
+    /// rather than polling with a timer
+    fn fetch_metrics(
+        state: &Self::State,
+        metrics: &prometheus::Registry,
+    ) -> Result<(), InitializationError>;
+
     /// Check the health of the connector.
     ///
     /// For example, this function should check that the connector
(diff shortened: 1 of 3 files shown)


src/connector.rs
```diff
@@ -180,6 +180,16 @@ pub trait Connector {
...
 
+    /// Update any metrics from the state
+    ///
+    /// Useful to query a connection pool
+    /// and use it to update metrics
+    /// rather than polling with a timer
+    fn fetch_metrics(
+        state: &Self::State,
+        metrics: &prometheus::Registry,
+    ) -> Result<(), InitializationError>;
```
danieljharvey, line 191: This should almost certainly not return an `InitializationError`, and instead a type of it's own.

src/connector.rs
```diff
@@ -180,6 +180,16 @@ pub trait Connector {
...
     ) -> Result<Self::State, InitializationError>;
 
+    /// Update any metrics from the state
+    ///
+    /// Useful to query a connection pool
+    /// and use it to update metrics
+    /// rather than polling with a timer
+    fn fetch_metrics(
+        state: &Self::State,
+        metrics: &prometheus::Registry,
```
paf31, line 190: I notice this isn't used currently, is there a use case in mind? All I can think of is dynamically registering metrics after `try_init_state`, but I'd be inclined to add that later if we need it.

src/connector.rs
paf31: Yes, please let's add a type for this.

paf31:
@danieljharvey To keep things moving along, I've modified the PR and I'll merge it once CI passes, but your postgres-ndc PR will need some slight changes now.

src/connector.rs
danieljharvey: That's a good point. I'd support removing it and adding it later if we need it.

PR #18, opened 2023-08-23
Pass OTEL_SERVICE_NAME to OpenTelemetry

danieljharvey:
Previously the OpenTelemetry service name was hardcoded as `ndc-hub`. This adds a serve command option that allows it to be set by the `OTEL_SERVICE_NAME` env var as described in the docs.

Here it is in action with Postgres NDC:

Screenshot 2023-08-23 at 12 07 33


rust-connector-sdk/src/default_main.rs
```diff
@@ -143,6 +145,11 @@ where
 fn init_tracing(serve_command: &ServeCommand) -> Result<(), Box<dyn Error>> {
     global::set_text_map_propagator(TraceContextPropagator::new());
 
+    let service_name = serve_command
+        .service_name
+        .clone()
+        .unwrap_or("ndc-hub".to_string());
```
SamirTalwar, line 151: Why not use `env!("CARGO_PKG_NAME")` here too?

danieljharvey: Aye, that's a good shout. Done!

PR #20, opened 2023-08-31
Test runner integration

paf31:
Depends on hasura/ndc-spec#25

Example output:

```
ndc_hub_example test --configuration <(echo null)

Capabilities
├ Fetching /capabilities ... ... OK
├ Validating capabilities ... OK
Schema
├ Fetching /schema ... OK
├ Validating schema ...
│ ├ object_types ... OK
│ ├ Collections ...
│ ├ Functions ...
│ │ ├ Procedures ...
Query
```


rust-connector-sdk/src/default_main.rs
```diff
@@ -391,3 +403,85 @@ where
...
+
+    let mut metrics = Registry::new();
+    let state = C::try_init_state(&configuration, &mut metrics)
+        .await
+        .unwrap();
+
+    let connector = ConnectorAdapter::<C> { configuration, state };
+    let results = ndc_test::test_connector(&test_configuration, &connector).await;
+
+    if !results.failures.is_empty() {
```
paf31, line 465: This block is copied verbatim from `ndc-test`, so we should find a way to extract this into a library function.

rust-connector-sdk/src/default_main.rs
```diff
@@ -391,3 +403,85 @@ where
...
+        println!();
+        println!(
+            "\x1b[1;31mFailed with {0} test failures:\x1b[22;0m",
```
sordina, line 493: Should use a term-colors crate or something in future. This is ok for now if we're already using this pattern.

rust-connector-sdk/src/default_main.rs
```diff
@@ -71,6 +74,14 @@ struct ServeConfigurationCommand {
     port: Option<String>,
 }
 
+#[derive(Clone, Parser)]
+struct TestCommand {
+    #[arg(long, value_name = "PORT", env = "PORT")]
+    seed: Option<String>,
+    #[arg(long, value_name = "CONFIGURATION_FILE", env = "CONFIGURATION_FILE")]
+    configuration: String,
```
sordina, line 87: Fails with serde parse error. 👍

rust-connector-sdk/src/default_main.rs
```diff
@@ -391,3 +403,85 @@ where
...
+struct ConnectorAdapter<C: Connector> {
+    configuration: C::Configuration,
+    state: C::State,
+}
+
+#[async_trait]
+impl<C: Connector> ndc_test::Connector for ConnectorAdapter<C>
+where
+    C::Configuration: Send + Sync + 'static,
+    C::State: Send + Sync + 'static,
```
sordina, line 438: Have to do this because the trait is in the spec. Maybe there's a utility to do this kind of adapter pattern. This is ok for now.

rust-connector-sdk/src/default_main.rs
```diff
@@ -391,3 +403,85 @@ where
...
+{
+    let test_configuration = ndc_test::TestConfiguration { seed: command.seed };
+
+    let configuration_json = std::fs::read_to_string(command.configuration).unwrap();
+    let raw_configuration =
+        serde_json::de::from_str::<C::RawConfiguration>(configuration_json.as_str()).unwrap();
+    let configuration = C::validate_raw_configuration(&raw_configuration)
+        .await
+        .unwrap();
```
sordina, line 477: Shouldn't there be a function that can be reused for this?

rust-connector-sdk/src/default_main.rs
```diff
@@ -391,3 +403,85 @@ where
...
+
+    let mut metrics = Registry::new();
+    let state = C::try_init_state(&configuration, &mut metrics)
+        .await
+        .unwrap();
+
+    let connector = ConnectorAdapter::<C> { configuration, state };
+    let results = ndc_test::test_connector(&test_configuration, &connector).await;
```
sordina, line 488: NDC Test Repo Reference.

sordina (approved):
Looks good. Up to you if you want to look into conditional inclusion flag and color improvements.

paf31:
@sordina This now uses the https://github.com/hasura/ndc-spec/pull/39 branch of spec, so will need to be updated again once that merged and tagged.

rust-connector-sdk/src/default_main.rs
paf31: I think we could share it between this function and the `serve` command potentially but I'll leave it out of this PR in the interest of time if you don't mind.

paf31:
Note you can now use `NO_COLOR=1` to turn off color in the console.

PR #24, opened 2023-09-11
Add a command that checks the server's health on the port provided.

SamirTalwar:
### Why

This allows us to add health checks to any `ndc-sdk`-based connector's Docker image without having to bundle `curl`.

### What

I added a `check-health` command that checks `<HOST>:<PORT>/health` and prints an error if it fails. It defaults to `localhost:8100`, and uses the `PORT` environment variable if it's set, for easier `exec`-ing inside a Docker container.

I also modified the `/health` endpoint to report the error given in the response body.

### How

It does this by checking that the HTTP status code is successful.

I also tidied up some other bits and bobs that got in my way.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -920,6 +920,7 @@ dependencies = [
  "tracing",
  "tracing-opentelemetry",
  "tracing-subscriber",
+ "url",
  "uuid",
 ]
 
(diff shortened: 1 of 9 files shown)


danieljharvey (approved):
This all seems very sensible to me.

PR #25, opened 2023-09-12
Add support for OpenTelemetry logs

danieljharvey:
Adds support for [OpenTelemetry Logs](https://opentelemetry.io/docs/specs/otel/logs/) (as opposed to the support we currently have for traces).

We want to use these for alerting important events, such as startup failures due to connection pool timeouts.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -906,7 +906,6 @@ dependencies = [
  "axum",
  "axum-macros",
  "clap",
- "log",
  "ndc-client",
  "opentelemetry",
  "opentelemetry-appender-log",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -802,9 +802,6 @@ name = "log"
 version = "0.4.19"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b06a4cde4c0f271a446782e3eff8de789548ce57dbc8eca9292c27f4a42004b4"
-dependencies = [
- "value-bag",
-]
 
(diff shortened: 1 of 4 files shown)


plcplc:
Looks plausible.

I tried reading the spec, but it's a bit hard for me to appreciate the difference between "Logging" and "Tracing".

It looks to me like the purpose of opentelemetry "logging" (as opposed to "tracing") is to integrate with existing logging frameworks. Is that happening here?

From https://github.com/open-telemetry/opentelemetry-rust#project-status:

> If you are starting fresh, then consider using [tracing](https://github.com/tokio-rs/tracing) as your logging API. It supports structured logging and is actively maintained.

A cursory look at that "tracing" package looks very similar to just opentelemetry tracing (with concepts such as "spans" and "traces"). It even has the name "tracing".

So I'm confused about what we're getting and why.

Is there a test (or just a manual way) that shows the resulting traces/logs?

danieljharvey:
I tested this with a Honeycomb instance.

Calling the reference agent `/health` endpoint produces this log:

```json
{
  "body": "Health check OK!",
  "event.domain": "database",
  "event.name": "health-check-success",
  "flags": "0",
  "library.name": "ndc_hub_example",
  "meta.signal_type": "log",
  "service.name": "ndc-hub",
  "service.version": "0.1.0",
  "severity": "info",
  "severity_code": 9
}
```

Screenshot 2023-09-13 at 12 16 31

PR #27, opened 2023-09-13
Adding SERVICE_TOKEN_SECRET to Rust SDK Serve command

sordina:
This is to be used in conjunction with the changes made to V3 engine that can send a pre-shared secret token via `Authorization: Bearer TOKEN` header.

If the `SERVICE_TOKEN_SECRET` env variable is set or the engine sends the `Authorization: Bearer SERVICE_TOKEN_SECRET` header, then the values must match.

If they do not match then the response will be StatusCode::UNAUTHORIZED.

If neither are set then the request proceeds as before.

### DIFF
--- a/rust-connector-sdk/src/default_main.rs
+++ b/rust-connector-sdk/src/default_main.rs
@@ -206,6 +206,7 @@ where
                 let auth_header = request.headers().get("Authorization")
                     .map(|v| v.clone());
 
+                // NOTE: The comparison should probably be more permissive to allow for whitespace, etc.
                 if auth_header == expected_auth_header { return Ok(()); }
                 Err((StatusCode::UNAUTHORIZED, "").into_response())
             })
--- a/rust-connector-sdk/src/default_main.rs
+++ b/rust-connector-sdk/src/default_main.rs
@@ -55,6 +55,8 @@ struct ServeCommand {
     otlp_endpoint: Option<String>, // NOTE: `tracing` crate uses `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` ENV variable, but we want to control the endpoint via CLI interface
     #[arg(long, value_name = "PORT", env = "PORT", default_value = "8100")]
     port: Port,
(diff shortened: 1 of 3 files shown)


rust-connector-sdk/src/default_main.rs
```diff
@@ -188,9 +192,25 @@ where
 
     let server_state = init_server_state::<C>(serve_command.configuration).await;
 
-    let router = create_router::<C>(server_state).layer(
-        TraceLayer::new_for_http().make_span_with(DefaultMakeSpan::default().level(Level::INFO)),
-    );
+    let expected_auth_header: Option<HeaderValue> = serve_command.service_token_secret.and_then(|service_token_secret| {
+        let expected_bearer = format!("Bearer {}", service_token_secret); // TODO
```
dmoverton, line 196: I don't think you need the "TODO" comment there any more.

sordina: Good point.

sordina: @dmoverton done

dmoverton (approved):
Looks good

rust-connector-sdk/src/default_main.rs
```diff
@@ -188,9 +192,25 @@ where
...
+            TraceLayer::new_for_http().make_span_with(DefaultMakeSpan::default().level(Level::INFO)),
+        ).layer(
+            ValidateRequestHeaderLayer::custom(move |request: &mut Request<Body>| {
+                // Validate the request
+                let auth_header = request.headers().get("Authorization")
+                    .map(|v| v.clone());
+
+                // NOTE: The comparison should probably be more permissive to allow for whitespace, etc.
+                if auth_header == expected_auth_header { return Ok(()); }
+                Err((StatusCode::UNAUTHORIZED, "").into_response())
```
paf31, line 211: I didn't notice this before but this should return an `ErrorResponse`.

sordina: Is there an example of how to do this?

sordina: https://github.com/hasura/ndc-hub/pull/35

PR #40, opened 2023-09-18
add v2 compatibility layer

BenoitRanque:
add `--enable_v2_compatibility` flag and env var. When enabled, a GDC server will be available under the `/v2` route.

Functionality is currently limited to mutations.
Some functionality is not supported, such as Array fields.

also had to conditionally remove sigterm handler so it would compile on windows

### DIFF
--- a/rust-connector-sdk/Cargo.toml
+++ b/rust-connector-sdk/Cargo.toml
@@ -48,5 +48,5 @@ tracing-subscriber = { version = "^0.3", default-features = false, features = [
 ] }
 url = "2.4.1"
 uuid = "^1.3.4"
-gdc_rust_types = { path = "../../gdc_rust_types" }
+gdc_rust_types = { git = "https://github.com/hasura/gdc_rust_types", rev = "bc57c40" }
 indexmap = "^1"
--- a/rust-connector-sdk/Cargo.toml
+++ b/rust-connector-sdk/Cargo.toml
@@ -49,4 +49,4 @@ tracing-subscriber = { version = "^0.3", default-features = false, features = [
 url = "2.4.1"
 uuid = "^1.3.4"
 gdc_rust_types = { path = "../../gdc_rust_types" }
-indexmap = "1.9.3"
+indexmap = "^1"
--- a/rust-connector-sdk/Cargo.toml
+++ b/rust-connector-sdk/Cargo.toml
@@ -18,10 +18,13 @@ axum-macros = "^0.3.7"
 clap = { version = "^4.3.9", features = ["derive", "env"] }
 ndc-client = { git = "http://github.com/hasura/ndc-spec.git", tag = "v0.1.0-rc.5" }
 ndc-test = { git = "http://github.com/hasura/ndc-spec.git", tag = "v0.1.0-rc.5" }
-opentelemetry = { version = "^0.20", features = [ "rt-tokio", "trace" ], default-features = false }
+opentelemetry = { version = "^0.20", features = [
+  "rt-tokio",
+  "trace",
+], default-features = false }
 opentelemetry_api = "^0.20.0"
 opentelemetry_sdk = "^0.20.0"
-opentelemetry-otlp = { version = "^0.13.0", features = [ "reqwest-client" ] }
+opentelemetry-otlp = { version = "^0.13.0", features = ["reqwest-client"] }
 opentelemetry-semantic-conventions = "^0.12.0"
 prometheus = "^0.13.3"
 reqwest = "^0.11.20"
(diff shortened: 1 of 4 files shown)


sordina:
Add to the PR description:

* Auth  configuration details for v2 projects
* Limitations
* How to enable when deploying with `connector create`
* How to enable for PG?

rust-connector-sdk/Cargo.toml
```diff
@@ -18,21 +18,35 @@ axum-macros = "^0.3.7"
...
+tracing-subscriber = { version = "^0.3", default-features = false, features = [
+  "ansi",
+  "env-filter",
+  "fmt",
+  "json",
+] }
 url = "2.4.1"
 uuid = "^1.3.4"
+gdc_rust_types = { path = "../../gdc_rust_types" }
+indexmap = "1.9.3"
```
sordina, line 52: Why are we pinning exact versions here?

rust-connector-sdk/src/default_main.rs
```diff
@@ -72,6 +76,8 @@ struct ServeCommand {
     service_token_secret: Option<String>,
     #[arg(long, value_name = "OTEL_SERVICE_NAME", env = "OTEL_SERVICE_NAME")]
     service_name: Option<String>,
+    #[arg(long, env = "ENABLE_V2_COMPATIBILITY")]
+    enable_v2_compatibility: bool,
```
sordina, line 80: 👍

rust-connector-sdk/src/default_main.rs
```diff
@@ -219,41 +225,17 @@ where
...
-                )
-                    .into_response())
-            },
-        ));
+    let router = if serve_command.enable_v2_compatibility {
+        let v2_router = create_v2_router(server_state, serve_command.service_token_secret.clone());
+        Router::new().merge(router).nest("/v2", v2_router)
+    } else {
+        router
+    };
```
sordina, line 239: It would be so good if we had some kind of regression tests. Cest la vie.

rust-connector-sdk/src/default_main.rs
```diff
@@ -317,20 +305,121 @@ where
...
+                                if provided_token_value == expected_token_value {
+                                    // if token set, config header must be present and include token
+                                    return Ok(());
+                                }
+                            }
+                        }
+                    }
+                } else {
+                    // if token not set, always authorize
+                    return Ok(());
```
sordina, line 401: Correct me if I'm wrong but I think this is a change in behaviour:

We should authorise only if:

* Neither the header or expected token are set
* or, Both the header and expected token match

and reject in all other cases.

This implementation authorises if:

* Expected token is not set, but header is set.

Should we allow this? Maybe but it is a change in behavoiur.

sordina (commented):
Looks pretty good IMO.

There's a fair bit being moved around so it would be great to get regression tests in place in future to help with this.

I think the behaviour has changed WRT expected auth token not being set, but I'm ok with the change if you are.

After you've tested on a v3 project and v2 project I think I can approve.

rust-connector-sdk/src/default_main.rs
BenoitRanque: Stricly speaking these are new endpoints so not a behavior change. But this is different from the v3 implementation.
I'll change it to match so we don't have to think about subtle behavior differences in the future.

rust-connector-sdk/src/default_main.rs
BenoitRanque: I'll do some old fashion "try it a few times" regression testing. Agreed I'd love to have more confidence this didn't break anything

rust-connector-sdk/src/default_main.rs
sordina: Good point.

BenoitRanque:
Did some testing with v3 engine, can confirm auth (still) works for v3 and we can query data just fine. Can also query data from v2 engine.

rust-connector-sdk/Cargo.toml
```diff
@@ -18,21 +18,35 @@ axum-macros = "^0.3.7"
...
-tracing-subscriber = { version = "^0.3", default-features = false, features = ["ansi", "env-filter", "fmt", "json"] }
+tracing-subscriber = { version = "^0.3", default-features = false, features = [
+  "ansi",
+  "env-filter",
+  "fmt",
+  "json",
+] }
 url = "2.4.1"
 uuid = "^1.3.4"
+gdc_rust_types = { path = "../../gdc_rust_types" }
```
SamirTalwar, line 51: Use a Git dependency ([here's an example from postgres-ndc](https://github.com/hasura/postgres-ndc/blob/bce8f232e75222d8f1f73f56ed76a9866549bed1/crates/ndc-postgres/Cargo.toml#L17)), not a path dependency to somewhere outside the repository.

BenoitRanque: 🤦 thanks and sorry about that

rust-connector-sdk/Cargo.toml
BenoitRanque: We don't need to. We do need to specify a version that is compatible with both `ndc-client` and `gdc_rust_types` which are the v3 and v2 api types respectively.
Changing to `^1` to match `ndc-client`

sordina (approved):
Approved.

PR #46, opened 2023-09-25
Add a health check for the configuration server that always passes.

SamirTalwar:
This allows us to use the `check-health` command for the configuration server too.


SamirTalwar:
This will let us remove the postgres-ndc Dockerfile that specifically runs the configuration server in favor of using the same Docker images that we already publish elsewhere.

danieljharvey (approved):

Issue #43, opened 2023-09-19
Attach STATUS_CODE_ERROR to traces for non-successful requests


sordina:
The routing trace layer should tag the span as an error if the request isn't successful. Currently:

```rust
        .layer(
            TraceLayer::new_for_http()
                .make_span_with(DefaultMakeSpan::default().level(Level::INFO)),
        )
```

But should attach the status code `STATUS_CODE_ERROR` type if the request isn't successful.

See: https://github.com/hasura/ndc-hub/blob/main/rust-connector-sdk/src/default_main.rs#L336

@karthikvt26

sordina:
@paf31 do you know how to do this?

PR #49, opened 2023-09-27
Implement trace propagation

danieljharvey:
This PR implements trace propagation, adds a `MakeSpan` trait that grab any `traceparent` and `tracestate` headers, and injecting them into span context. The result is that requests from `v3-engine` show up in NDC agents as follows:

Screenshot 2023-09-29 at 12 52 35

### DIFF
--- a/justfile
+++ b/justfile
@@ -21,6 +21,3 @@ lint *FLAGS:
 
 lint-apply *FLAGS:
   cargo clippy --fix {{FLAGS}}
-
-
-
--- a/justfile
+++ b/justfile
@@ -1,5 +1,26 @@
+# re-build on code changes, and run the reference agent each time a build is
+# successful
 dev:
   cargo watch \
+    -x test \
     -x 'run --bin ndc_hub_example \
     -- serve --configuration <(echo 'null') \
     --otlp-endpoint http://localhost:4317'
+
+# reformat everything
+format:
+  cargo fmt --all
+
+# is everything formatted?
+format-check:
+  cargo fmt --all --check
+
+# run `clippy` linter
+lint *FLAGS:
+  cargo clippy {{FLAGS}}
+
+lint-apply *FLAGS:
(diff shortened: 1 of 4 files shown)


rust-connector-sdk/src/default_main.rs
```diff
@@ -331,10 +333,7 @@ where
         });
 
     router
-        .layer(
-            TraceLayer::new_for_http()
-                .make_span_with(DefaultMakeSpan::default().level(Level::INFO)),
```
danieljharvey, line 336: Removing this `make_span_with` means we default to the default `TraceLayer` implementation, which makes all error responses into error traces rather than always `INFO`, solving #43 

Info: https://docs.rs/tower-http/latest/tower_http/trace/struct.TraceLayer.html#method.new_for_http

rust-connector-sdk/src/default_main.rs
```diff
@@ -359,6 +358,19 @@ where
         ))
 }
 
+#[derive(Clone, Copy, Debug)]
+pub struct PropagateParentTraces;
+
+impl<B> MakeSpan<B> for PropagateParentTraces {
+    fn make_span(&mut self, req: &Request<B>) -> Span {
```
danieljharvey, line 365: This grabs the W3 trace header (ie, `traceheader` and `tracestate`) and adds them to the context of any spans we create, meaning we can follow between services.

justfile
```diff
@@ -0,0 +1,23 @@
+# re-build on code changes, and run the reference agent each time a build is
```
danieljharvey, line 1: A few quality-of-life `just` commands. Happy to move these into a separate PR if needed.

danieljharvey:
Just realised there is already an implementation in `ndc-multitenant`, will copy that instead for the sake of consistency: https://github.com/hasura/ndc-hub/pull/49

danieljharvey:
Closing in favour of #51

PR #55, opened 2023-10-09
Wrap state in `Arc` to avoid cloning it.

SamirTalwar:
axum clones state on every request, making it very expensive. We can avoid this by wrapping it in an `Arc`, which means we just clone the reference. [This is encouraged by the axum documentation.](https://docs.rs/axum/latest/axum/index.html#sharing-state-with-handlers)

With this change, we no longer need the connector, configuration, or state to be cloneable.

As state is mostly read-only, this seems to work without further changes, except for explicitly cloning the metrics registry too. (The metrics registry is itself cheap to clone, so no issues there.)

### DIFF
--- a/rust-connector-sdk/src/default_main.rs
+++ b/rust-connector-sdk/src/default_main.rs
@@ -114,7 +114,7 @@ struct CheckHealthCommand {
 
 type Port = u16;
 
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct ServerState<C: Connector> {
     configuration: C::Configuration,
     state: C::State,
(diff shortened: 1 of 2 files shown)


SamirTalwar:
@danieljharvey, on my machine, I see about a 10% improvement in benchmarks on ndc-postgres. Let's see if they replicate in CI once this is merged.

paf31 (commented):
To note: originally I implemented this as-is because I figured the connectors could implement `Clone` using `Arc` on the bits of state they wanted to share. I didn't want to choose `Arc` as the sharing mechanism for all of the state on the behalf of every connector implementer.

SamirTalwar:
Gonna close this in favor of doing this in individual connectors.

PR #56, opened 2023-10-09
Move configuration when we don't need to hold onto it.

SamirTalwar:
There are certain functions in the `Connector` trait that use the raw configuration, and accept a reference. We don't actually need to keep the configuration around in the SDK so we can pass by value here, which means that the relevant functions can own the data if they require it.

Making these input parameters values rather than references allows us to avoid cloning in other locations later.

### DIFF
--- a/rust-connector-sdk/src/connector.rs
+++ b/rust-connector-sdk/src/connector.rs
@@ -192,13 +192,13 @@ pub trait Connector {
     fn make_empty_configuration() -> Self::RawConfiguration;
 
     async fn update_configuration(
-        config: &Self::RawConfiguration,
+        config: Self::RawConfiguration,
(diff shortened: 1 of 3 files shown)


SamirTalwar:
This is a breaking change. I assume that as we're on v0.1.0, we're fine with this, but please shout if we should avoid them.

plcplc (commented):
I think this is great and LGTM.

I'm holding back on approving until the potential versioning issue has had some time for commentary.

soupi (approved):
LGTM.

SamirTalwar:
A cursory scan of the recently-merged PRs tells me we're alright making breaking changes for now, so I'm going to merge this.

Thanks all!

PR #58, opened 2023-10-10
Introduce a "fast path" for returning already-serialized JSON.

SamirTalwar:
This introduces a new enum, `JsonResponse`, which is designed to supplant `axum::Json`. It implements `axum::IntoResponse`.

It has two variants:

1. `Value`, which behaves the same as `axum::Json`.
2. `Serialized`, which allows the connector to provide already-serialized JSON in the form of a `Bytes` value.

The latter can be used to construct the JSON directly in the database and avoid deserialization and re-serialization.

This is a breaking change, as at the very least, the connector implementations will now need to wrap responses in `JsonResponse::Value`. Hopefully it's not too much of a big deal.

There is a helper function, `JsonResponse::into_value`, used to deserialize the serialized bytes in certain situations. This is because the v2 compatibility layer and the connector used for ndc-test both require transforming the data, and so cannot work directly with bytes. We should not call this when running the connector normally, and it is not available outside this crate.

### DIFF
--- a/rust-connector-sdk/src/connector.rs
+++ b/rust-connector-sdk/src/connector.rs
@@ -3,6 +3,9 @@ use ndc_client::models;
 use serde::Serialize;
 use std::error::Error;
 use thiserror::Error;
+
+use crate::json_response::JsonResponse;
+
 pub mod example;
 
 /// Errors which occur when trying to validate connector
@@ -35,7 +38,7 @@ pub enum KeyOrIndex {
 #[derive(Debug, Error)]
 pub enum UpdateConfigurationError {
     #[error("error validating configuration: {0}")]
-    Other(Box<dyn Error + Send + Sync>),
+    Other(#[from] Box<dyn Error + Send + Sync>),
 }
 
 /// Errors which occur when trying to initialize connector
@@ -45,7 +48,7 @@ pub enum UpdateConfigurationError {
 #[derive(Debug, Error)]
 pub enum InitializationError {
     #[error("error initializing connector state: {0}")]
-    Other(Box<dyn Error + Send + Sync>),
+    Other(#[from] Box<dyn Error + Send + Sync>),
 }
 
 /// Errors which occur when trying to update metrics.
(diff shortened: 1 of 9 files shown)


rust-connector-sdk/src/connector.rs
```diff
@@ -35,7 +38,7 @@ pub enum KeyOrIndex {
 #[derive(Debug, Error)]
 pub enum UpdateConfigurationError {
     #[error("error validating configuration: {0}")]
-    Other(Box<dyn Error + Send + Sync>),
+    Other(#[from] Box<dyn Error + Send + Sync>),
```
soupi, line 41: What does `#[from]` do?

rust-connector-sdk/src/json_response.rs
```diff
@@ -0,0 +1,116 @@
+use axum::response::IntoResponse;
+use bytes::Bytes;
+use http::{header, HeaderValue};
+
+/// Represents a response value that will be serialized to JSON.
+///
+/// The value may be of a type that implements `serde::Serialize`, or it may be
+/// a contiguous sequence of bytes, which are _assumed_ to be valid JSON.
+#[derive(Debug, Clone)]
+pub enum JsonResponse<A: serde::Serialize + (for<'de> serde::Deserialize<'de>)> {
```
hallettj, line 10: I don't think you need these trait bounds here, unless it makes errors easier to track down

hallettj (approved):
It looks good to me.

I think it would be helpful to include this impl so that users can write `.into()` instead of wrapping values in `JsonResponse::Value( )`:

```rust
impl<A> From<A> for JsonResponse<A> {
    fn from(value: A) -> Self {
        JsonResponse::Value(value)
    }
}
```

It would by nice if we could do something like this, but I don't see any way to make this work without introducing an associated type:

```rust
fn get_capabilities() -> impl Into<JsonResponse<models::CapabilitiesResponse>>;
```

But that might work in a future Rust version.

rust-connector-sdk/src/json_response.rs
SamirTalwar: Ah, probably not, but I would prefer to get the warnings earlier.

SamirTalwar:
Unfortunately, `impl<A> From<A> for JsonResponse<A>` is not allowed because we don't own `A`, and Rust is not as clever as Haskell.

SamirTalwar:
Good idea, adding `impl From<A>`; done.

rust-connector-sdk/src/json_response.rs
SamirTalwar: Removed this after all because it was replicating everywhere.

soupi (approved):
lgtm.

PR #62, opened 2023-10-18
Add SecretValue, update for latest ndc-test

paf31:
This ought to have been two PRs, but these will go out together in the nest `ndc-spec` tag anyway.

- Update for latest `ndc-test` - add `snapshot-dir` argument and new `replay` command to `default_main`
- Add `SecretValue` type here instead of in `ndc-spec`

### DIFF
--- a/rust-connector-sdk/src/secret.rs
+++ b/rust-connector-sdk/src/secret.rs
@@ -1,0 +1,70 @@
+use schemars::{self, schema::Schema, JsonSchema};
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+#[schemars(title = "SecretValue")]
+/// Either a literal string or a reference to a Hasura secret
+pub enum SecretValueImpl {
+    Value(String),
+    StringValueFromSecret(String),
+}
+
+/// Use this type to refer to secret strings within a
+/// connector's configuration types. For example, a connection
+/// string which might contain a password should be configured
+/// using this type.
+///
+/// This marker type indicates that a value should be configured
+/// from secrets drawn from a secret store.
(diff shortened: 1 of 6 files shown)


rust-connector-sdk/src/secret.rs
```diff
@@ -0,0 +1,70 @@
...
+        if let Schema::Object(o) = &mut s {
+            if let Some(m) = &mut o.metadata {
+                m.id = Some(Self::schema_id().into());
+            }
+        }
+        s
+    }
+
+    fn schema_id() -> std::borrow::Cow<'static, str> {
+        "hasura.io/SecretValue".into()
```
abhinav-hasura, line 38: Looks like this needs to be a proper URI reference (but not necessary network reachable) according to the JSONSchema spec.

How about: "https://hasura.io/jsonschemas/SecretValue"?

abhinav-hasura (approved):

rust-connector-sdk/src/secret.rs
```diff
@@ -0,0 +1,70 @@
+use schemars::{self, schema::Schema, JsonSchema};
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug, Clone, JsonSchema)]
+#[serde(rename_all = "camelCase")]
+#[schemars(title = "SecretValue")]
+/// Either a literal string or a reference to a Hasura secret
+pub enum SecretValueImpl {
+    Value(String),
+    StringValueFromSecret(String),
```
SamirTalwar, line 10: Will this ever be constructed, or is this purely for generating the JSON schema?

abhinav-hasura: Right now I don't think anyone constructs this, but at some point I can see MBS constructing this when deserializing a secretable value.

paf31:
Will split this from the `ndc-test` change and make a separate PR.

PR #66, opened 2023-10-26
Add status code metrics

soupi:
## What

https://hasurahq.atlassian.net/browse/NDAT-1006

We'd like to track the status codes returned by a connector in metrics.

## How

From the metrics registry created by ndc-sdk, create a metrics struct that will contain a counter for each status code we [want to track](https://hasura.github.io/ndc-spec/specification/error-handling.html), and increment those on each relevant API call (query, explain, etc.). We also include group buckets for 1xx, 2xx, etc.

An alternative to this approach would be to create a middleware that would count all responses, though these will also include health checks, missing api calls, metrics api calls, etc. Which I thought would muddle the metrics we actually care about.

We also add a new function to the Connector trait - `fn connector_name() -> String`, so we can use it to create metrics names for the connector.

### DIFF
--- a/rust-connector-sdk/src/connector.rs
+++ b/rust-connector-sdk/src/connector.rs
@@ -192,6 +192,8 @@ pub trait Connector {
     /// The type of unserializable state
     type State;
 
+    fn connector_name() -> String;
+
     fn make_empty_configuration() -> Self::RawConfiguration;
 
     async fn update_configuration(
--- a/rust-connector-sdk/src/metrics.rs
+++ b/rust-connector-sdk/src/metrics.rs
@@ -50,67 +50,67 @@ impl Metrics {
 
         let total_200 = add_int_counter_metric(
             metrics_registry,
-            format!("{}_status_code_200_total_count", connector_name).as_str(),
+            format!("hasura_{}_status_code_200_total_count", connector_name).as_str(),
             "Total number of 200 status codes returned.",
         )?;
 
         let total_400 = add_int_counter_metric(
             metrics_registry,
-            format!("{}_status_code_400_total_count", connector_name).as_str(),
+            format!("hasura_{}_status_code_400_total_count", connector_name).as_str(),
             "Total number of 400 status codes returned.",
         )?;
 
         let total_403 = add_int_counter_metric(
             metrics_registry,
-            format!("{}_status_code_403_total_count", connector_name).as_str(),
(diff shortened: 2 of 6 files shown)


rust-connector-sdk/src/metrics.rs
```diff
@@ -0,0 +1,145 @@
...
+}
+
+impl Metrics {
+    /// Set up counters and gauges used to produce Prometheus metrics
+    pub fn initialize(
+        metrics_registry: &mut prometheus::Registry,
+    ) -> Result<Self, prometheus::Error> {
+        let total_200 = add_int_counter_metric(
+            metrics_registry,
+            "status_code_200_total_count",
```
paf31, line 32: These names need to be reserved if we're claiming them here. The only reserved names in the spec right now are `hasura_*`, but those are reserved for future use.

SamirTalwar:
I suggest doing both for now; they'll be useful in different situations, especially as we're not 100% familiar with the metrics and alerting tools that we'll be using. We can remove one if it's clear it's superfluous.

rust-connector-sdk/src/metrics.rs
danieljharvey: All the other metrics we emit in `ndc_posgres` are prefixed like `postgres_ndc_rest_of_metric_name`. Could we do that here? I guess maybe we'd want to add the prefix to the `Connector` so each agent specifies what it will use?

rust-connector-sdk/src/connector.rs
```diff
@@ -192,6 +192,8 @@ pub trait Connector {
     /// The type of unserializable state
     type State;
 
+    fn connector_name() -> String;
```
soupi, line 195: Introducing a new static method to the Connector trait.

rust-connector-sdk/src/metrics.rs
```diff
@@ -0,0 +1,239 @@
...
+                    Some(c)
+                } else {
+                    None
+                }
+            })
+            .collect();
+
+        let total_200 = add_int_counter_metric(
+            metrics_registry,
+            format!("hasura_{}_status_code_200_total_count", connector_name).as_str(),
```
SamirTalwar, line 53: Note that `.as_str()` can also be done using `&`.

I don't have a particular preference for one over the other.

SamirTalwar (approved):

PR #72, opened 2023-11-02
return extraction errors as json ErrorResponse

soupi:
## What

[We'd like](https://hasurahq.slack.com/archives/C04NS5JCD8A/p1697618175968319?thread_ts=1697551776.418919&cid=C04NS5JCD8A) to return a json [ErrorResponse](https://hasura.github.io/ndc-spec/reference/types.html#errorresponse) on extract errors instead of plain text error messages.

With this PR, we'll return (for example):

```json
{
  "message": "Parse error",
  "details": "Failed to deserialize the JSON body into the target type: query.where.type: unknown variant `inary_comparison_operator`, expected one of `and`, `or`, `not`, `unary_comparison_operator`, `binary_comparison_operator`, `binary_array_comparison_operator`, `exists` at line 1 column 210"
}
```

## How

We use [WithRejection](https://docs.rs/axum-extra/latest/axum_extra/extract/struct.WithRejection.html) to apply a transformation to the errors that the normal `Json` extractor emits, and wrap them in a json object.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -162,6 +162,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "axum-extra"
+version = "0.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "4ab90e7b70bea63a153137162affb6a0bce26b584c24a4c7885509783e2cf30b"
+dependencies = [
+ "axum",
+ "axum-core",
+ "bytes",
+ "futures-util",
+ "http",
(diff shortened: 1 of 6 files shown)


danieljharvey (approved):
Assuming this doesn't change the types of the `Connector` trait (which I'm 99% sure this doesn't) then this looks nice!

soupi:
@danieljharvey no change to the trait. These are the wrappers that call the connector methods and then convert the results to responses.

PR #73, opened 2023-11-06
update ndc spec dep

paritosh-08:
WIP PR

We need to add the tag instead of rev. So, once the ndc-spec is tagged, we will update the PR and mark it ready for review.


paritosh-08:
Closing this PR in favour of https://github.com/hasura/ndc-hub/pull/77

PR #76, opened 2023-11-07
Log errors deserializing request bodies

plcplc:
We want to log when we cannot fundamentally make sense of the requests we receive.

### DIFF
--- a/rust-connector-sdk/src/json_rejection.rs
+++ b/rust-connector-sdk/src/json_rejection.rs
@@ -1,29 +1,33 @@
 //! We want errors returned from failed json extractors to be formatted as json as well.
 
 use axum::extract;
-use axum::{http::StatusCode, response::IntoResponse};
+use axum::response::IntoResponse;
 use ndc_client::models;
 
-pub struct JsonRejection {
-    error: models::ErrorResponse,
-    status: StatusCode,
-}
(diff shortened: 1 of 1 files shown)


rust-connector-sdk/src/json_rejection.rs
```diff
@@ -1,29 +1,33 @@
...
+            name = "Unable to deserialize request body",
+            body = %rejection.body_text(),
+            error = true,
+        );
+        let error = models::ErrorResponse {
+            message: "Parse error".to_string(),
+            details: serde_json::Value::String(rejection.body_text()),
+        };
+        let payload = serde_json::to_value(error).unwrap();
+        (rejection.status(), extract::Json(payload)).into_response()
```
SamirTalwar, line 31: This seems a little bananas to me. Why are we using an extractor in a response?

I realize you didn't write this, but it does not seem correct.

Let's look into this separately.

SamirTalwar (approved):

rust-connector-sdk/src/json_rejection.rs
plcplc: It turns out that this makes sense to an extent, since it appears to be Axum's way to communicate/build up IntoResponse trait implementations from smaller parts: https://docs.rs/axum/0.6.20/axum/response/trait.IntoResponse.html

PR #77, opened 2023-11-14
Upgrade ndc-spec and gdc_rust_types.

SamirTalwar:
Notable changes:

- mutations were removed
- `source_collection_or_type` was removed
- `indexmap` was upgraded to version 2

### DIFF
--- a/rust-connector-sdk/Cargo.toml
+++ b/rust-connector-sdk/Cargo.toml
@@ -12,9 +12,9 @@ name = "ndc_hub_example"
 path = "bin/main.rs"
 
 [dependencies]
-gdc_rust_types = { git = "https://github.com/hasura/gdc_rust_types.git", rev = "bc57c40" }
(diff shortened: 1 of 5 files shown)


rust-connector-sdk/Cargo.toml
```diff
@@ -23,27 +23,27 @@ base64 = "0.21.5"
 bytes = "1.5.0"
 clap = { version = "^4.4.7", features = ["derive", "env"] }
 http = "^0.2"
-indexmap = "^1"
+indexmap = "2"
```
paf31, line 26: Can we relax this to `^2`?

SamirTalwar: They're identical.

> **Caret requirements** are an alternative syntax for the default strategy, `^1.2.3` is exactly equivalent to `1.2.3`.

— [Caret requirements](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements)

danieljharvey (approved):

PR #78, opened 2023-11-15
return capabilities in /validate

abhinav-hasura:
We want to start respecting connector capabilities in the engine, so we need them when building metadata in MBS using /validate.


paf31:
Let's update `spec` to `rc.11` too, and then you'll have the latest capabilities structure?

abhinav-hasura:
> Let's update `spec` to `rc.11` too, and then you'll have the latest capabilities structure?

Done

paf31 (approved):

PR #79, opened 2023-11-21
Only set the span parent ID if it is valid.

SamirTalwar:
If there is no parent span ID, we something nonsensical, so we need to validate it.

Yes, this is hilarious.

### DIFF
--- a/rust-connector-sdk/src/tracing.rs
+++ b/rust-connector-sdk/src/tracing.rs
@@ -85,7 +85,7 @@ pub fn make_span(request: &Request<Body>) -> Span {
     let parent_context = global::get_text_map_propagator(|propagator| {
         propagator.extract(&HeaderExtractor(request.headers()))
     });
-    // if there is no parent span ID, we something nonsensical, so we need to validate it
+    // if there is no parent span ID, we get something nonsensical, so we need to validate it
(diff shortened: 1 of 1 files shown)


SamirTalwar:
Before:

before, direct

After:

after, direct

Note that this only affects direct requests. Via the engine, there is no change. This is therefore mostly useful in testing, not when running behind an engine.

rust-connector-sdk/src/tracing.rs
```diff
@@ -83,7 +85,13 @@ pub fn make_span(request: &Request<Body>) -> Span {
     let parent_context = global::get_text_map_propagator(|propagator| {
         propagator.extract(&HeaderExtractor(request.headers()))
     });
-    span.set_parent(parent_context);
+    // if there is no parent span ID, we something nonsensical, so we need to validate it
```
plcplc, line 88: This not quite grammatical..

plcplc (approved):

rust-connector-sdk/src/tracing.rs
SamirTalwar:
```suggestion
    // if there is no parent span ID, we get something nonsensical, so we need to validate it
```

PR #85, opened 2024-01-04
Add an 'UnprocessableContent' error

soupi:
## What

We want to be able to return 422 error codes on cases such as the following query:

```graphql
query MyQuery {
  ArtistByArtistId(ArtistId: "") {
    Name
  }
}
```

Where the user provides an invalid value - `ArtistId` is expected to be an integer, but the user provided an empty string.

## How

To do that we add a new error variant for QueryError, ExplainError and MutationError which will be mapped to that status code 422.

### DIFF
--- a/rust-connector-sdk/src/connector.rs
+++ b/rust-connector-sdk/src/connector.rs
@@ -89,7 +89,7 @@ pub enum QueryError {
     /// an error with the client.
     #[error("invalid request: {0}")]
     InvalidRequest(String),
-    /// The request was well formed but was unabled to be
+    /// The request was well formed but was unable to be
     /// followed due to semantic errors. This indicates
     /// an error with the client.
     #[error("unprocessable content: {0}")]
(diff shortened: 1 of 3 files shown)


rust-connector-sdk/src/connector.rs
```diff
@@ -89,6 +89,11 @@ pub enum QueryError {
     /// an error with the client.
     #[error("invalid request: {0}")]
     InvalidRequest(String),
+    /// The request was well formed but was unabled to be
```
paf31, line 92:
```suggestion
    /// The request was well formed but was unable to be
```

typo

rust-connector-sdk/src/connector.rs
```diff
@@ -108,6 +113,11 @@ pub enum ExplainError {
...
```
paf31, line 116:
```suggestion
    /// The request was well formed but was unable to be
```

paf31 (approved):
Looks good to me, thanks!

PR #89, opened 2024-01-28
RFC: packaging spec

paf31:
[Rendered](https://github.com/hasura/ndc-hub/blob/rfc/packaging-spec/rfcs/0001-packaging.md)

- [ ] Notes on `rust-connector-sdk`
- [x] Notes on Hub manifest changes

### DIFF
--- a/rfcs/0001-packaging.md
+++ b/rfcs/0001-packaging.md
@@ -1,4 +1,4 @@
-# Connector Packaging
+# Connector Packaging and Hub Connector Definition
 
 ## Purpose
 
@@ -13,62 +13,46 @@ Hasura Hub data connectors are packaged as Docker images which follow the [deplo
 - Connectors that do not require a build step can be described using a name and version, which should correspond to an entry in the connector hub registry.
 - Connectors that require a build step (performed over some build inputs before they can execute correctly) can be described using a Dockerfile bundled with any additional build inputs.
 
-### Connector Configuration
-
-_TODO_: describe the layout of files on disk for prebuilt connectors, and for connectors which require a build step. Is this the same as the layout described in "Hasura Hub Connector Definition" below?
-
 ### Hasura Hub Connector Definition
 In order for the Hasura tooling to understand a connector, know how to interact with it and know how it is packaged, a connector will need a definition that contains this information.
 
-The connector definition package is a .tgz (tar file, gzipped) that contains the following files:
+The connector definition is a file structure that contains the following files and directories:
 
 ```
 /
-  .hasura/
-    connector-metadata.json # See ConnectorMetadataDefinition type
-    docker-compose.yaml # If DockerComposeWatchMode is used
-    
-    ### Start: If Docker packaging requires a build step
+  connector-metadata.json # See ConnectorMetadataDefinition type
+  
+  docker-compose.yaml
+
+  build-files/ # Connector specific build/configuration files
+    src/
+      functions.ts
+    package.json
+    package-lock.json
+    tsconfig.json
+
+  docker/ # If ManagedDockerBuildPackaging is used
     .dockerignore
     Dockerfile
-    ### End: If Docker packaging requires a build step
-    
-  ### Start: Connector specific configuration files
-  src/
-    functions.ts
-  package.json
-  package-lock.json
-  tsconfig.json
-  ### End: Connector specific configuration files
-
 ```
 
 ```typescript
 type ConnectorMetadataDefinition = {
   packagingDefinition: PackagingDefinition
   supportedEnvironmentVariables: EnvironmentVariableDefinition[]
+  cliPlugin?: CliPluginDefinition
 }
 
-type PackagingDefinition = PrebuiltDockerImagePackaging | DockerBuildPackaging
+type PackagingDefinition = PrebuiltDockerImagePackaging | ManagedDockerBuildPackaging
 
 type PrebuiltDockerImagePackaging = {
   type: "PrebuiltDockerImage"
   dockerImage: string // eg "hasura/postgres-data-connector:1.0.0"
 }
 
-type DockerBuildPackaging = {
-  type: "DockerBuild"
-  watchMode: WatchMode
-}
-
-type WatchMode = DockerRebuildWatchMode | DockerComposeWatchMode | ShellCommandWatchMode
-
-type DockerRebuildWatchMode = {
-  type: "DockerRebuild"
-}
-
-type DockerComposeWatchMode = {
-  type: "DockerCompose"
+type ManagedDockerBuildPackaging = {
+  type: "ManagedDockerBuild"
+  additionalWatchModes: ShellCommandWatchMode[]
 }
 
 type ShellCommandWatchMode = {
@@ -81,44 +65,47 @@ type EnvironmentVariableDefinition = {
   description: string
   defaultValue?: string
 }
+
+type CliPluginDefinition = {
+  pluginPackageUrl: string // https://url/to/download/plugin/to/install.tgz
+  pluginPackageSHA256: string // To ensure plugin integrity and also to identify which version to use out of those installed
+  watchSubcommand?: string // If the connector wants to add to the watch mode, it can define what subcommand to use "watch --whatever --flags"
+}
 ```
 
-The `.hasura/connector-metadata.json` contains JSON that describes:
+The `connector-metadata.json` contains JSON that describes:
 - The environment variables the connector supports to configure it (`supportedEnvironmentVariables`)
-- The packaging definition, which can be either `PrebuiltDockerImagePackaging` (a connector that does not require a build step), or  `DockerBuildPackaging` (a connector that requires a build step).
+- The packaging definition, which can be either `PrebuiltDockerImagePackaging` (a connector that does not require a build step), or  `ManagedDockerBuildPackaging` (a connector that requires a build step).
   - `PrebuiltDockerImagePackaging` defines the prebuilt `dockerImage` used to run the connector (`dockerImage`)
-  - If `DockerBuildPackaging` is used, a Dockerfile must be in the `.hasura` directory (and optionally, a `.dockerignore`). It will be used to build the connector. A `watchMode` must be specified, to tell the Hasura tooling how to perform watching for this connector (see [Watch Mode section](#watch-mode)).
+  - If `ManagedDockerBuildPackaging` is used, a Dockerfile must be in the `.hasura` directory (and optionally, a `.dockerignore`). It will be used to build the connector.
+- An optional `CLIPluginDefinition` that describes where to acquire the CLI plugin for this connector that can be used to enhance watch mode with connector-specific configuration updates
 
-#### Watch Mode
+### Watch Mode
 When developing locally using the connector, the user may utilize the connector in a "watch mode" that automatically reloads the connector with new configuration as they change the configuration.
 
+Every connector must specify a `docker-compose.yaml` in their connector definition that defines how to watch the container.
+
 #### Connectors that do not require a build step (`PrebuiltDockerImagePackaging`)
-If any configuration file changes, the connector image is restarted with the new version of the configuration file mounted.
-
-#### Connectors that require a build step (`DockerBuildPackaging`)
-There are three options a connector can choose from in their `connector-metadata.json`:
-
-##### `DockerRebuildWatchMode`
-In this mode, if any configuration file changes, the Docker container is rebuilt and restarted. This is the simplest mode, but the least performant, unless your Docker builds are extremely quick.
-
-##### `DockerComposeWatchMode`
-In this mode, connectors can provide a `.hasura/docker-compose.yaml` file. This file should:
-
-- Define a single service, which represents the connector 
-- Build that service using the `.hasura/Dockerfile`
-- Define a port mapping that uses the environment variable `PORT` for the host port and 8080 for the container port. This allows the watch tooling to control the port used by the connector and exposes the port on the host.
-- Provide a [docker compose watch configuration](https://docs.docker.com/compose/compose-file/develop/#watch) that defines how to watch the configuration files and how to react to different files changing. It can react by causing a container rebuild and restart, or by copying the new file inside the existing container and optionally restarting the container. This allows the connector to optimise its hot reload capability by only performing a rebuild where necessary.
-
-For example, a NodeJS-based connector that needs to perform an npm package restore as a part of its Docker build may define the following `.hasura/docker.compose.yaml`:
+The `docker-compose.yaml` should contain something like this:
 
 ```yaml
 services:
   connector:
-    build: 
-      context: ../ # The build context is the parent directory of .hasura/
-      dockerfile: ./Dockerfile
-    ports:
-      - ${PORT}:8080 # Required so that the watch tooling can control the port used
+    develop:
+      watch:
+        - path: ./
+          target: /etc/connector
+          action: sync+restart
+```
+
+The `connector` service must be defined, and note that no `image` or `build` property is defined. The CLI tooling will [extend](https://docs.docker.com/compose/compose-file/05-services/#extends) from this to add the necessary pre-built image. This exists solely to provide the watch configuration.
+
+#### Connectors that require a build step (`ManagedDockerBuildPackaging`)
+The `docker-compose.yaml` should contain something like this (specific to the watch requirements of the connector):
+
+```yaml
+services:
+  connector:
     develop:
       watch:
         # Rebuild the container if a new package restore is required because package[-lock].json changed
@@ -132,7 +119,11 @@ services:
         - path: ./
           target: /etc/connector
           action: sync+restart
-  ```
+```
+
+The `connector` service must be defined, and note that no `image` or `build` property is defined. The CLI tooling will [extend](https://docs.docker.com/compose/compose-file/05-services/#extends) from this to add the necessary `build` property based on the build context from `ConnectorManifest` and the Dockerfile included in the Connector Definition. This exists solely to provide the watch configuration.
+
+`ManagedDockerBuildPackaging` connectors can also optionally provide additional watch modes. This is intended to support connectors that can run using local developer machine tooling (for example, local NodeJS for debugging purposes). These additional watch modes may be chosen by the user to run _instead_ of the docker compose watch-based watch mode. They are configured via the `additionalWatchModes` property, which currently only supports `ShellCommandWatchMode`.
 
 ##### `ShellCommandWatchMode`
 In this mode, connectors can provide some native way of performing a hot-reloading watch mode. Whatever this custom method is, it will be started by running the shell command defined. This can be useful to allow local tooling to be used to run the connector and perform hot reloading.
@@ -144,6 +135,61 @@ In this mode, connectors can provide some native way of performing a hot-reloadi
 
 For example, for the NodeJS Lambda Connector, it could set the watch shell command to `npm run watch`, which would run the connector and activate its built-in hot-reloading functionality.
 
+#### CLI Plugin Watch Mode
+If connectors need to extend the watch process with their own functionality, they can implement a CLI plugin that contains a `watch` subcommand. If specified in the `connector-metadata.json` the tooling will use it in addition to existing watch functionality.
+
+An example use case for this would be for a Postgres connector, where you may want to watch the database for schema changes, and when they occur, update the schema introspection configuration file in the connector's build directory with the latest schema details. The docker compose watch would then reload the connector with the updated introspection configuration file.
+
+### Connector Layout in Hasura Project
+When a new connector is added to a Hasura project using `hasura3 add ConnectorManifest --hub hasura/nodejs-lambda:1.0`, the CLI acquires the Hasura Hub Connector Definition for the specified Hub Connector. It then places this inside the `~/.hasura/hub-connectors/` directory. This is done to keep non-user editable files out of the user's source tree.
+
+The CLI then puts the `build-files/` into the build directory for that connector and adds a `.build.hml` file with the `ConnectorManifest` metadata object.
+
+```
+# A copy of the `Hasura Hub Connector Definition` for a particular hub connector version
+~/.hasura/hub-connectors/hasura/nodejs-lambda/1.0/ 
+  connector-metadata.json
+  docker-compose.yaml
+  build-files/
+    src/
+      functions.ts
+    package.json
+    package-lock.json
+    tsconfig.json
+  docker/
+    .dockerignore
+    Dockerfile
+
+project/
+  subgraphs/default/dataconnectors/
+    my-dataconnector/
+      build/
+        ### Start: from `build-files/`
+        src/
+          functions.ts
+        package.json
+        package-lock.json
+        tsconfig.json
+        ### End: from `build-files/`
+
+        my-dataconnector.build.hml # Connector manifest, see below
+```
+
+Here's an example of what `my-dataconnector.build.hml` would contain (`ConnectorManifest` is not specified by this RFC):
+```yaml
+kind: ConnectorManifest
+definition:
+  name: my-dataconnector
+  type: local
+  hub-connector: hasura/nodejs-lambda:1.0 # This is used to find the Hasura Hub Connector Definition in ~/.hasura/hub-connectors/
+  tunnel: true
+  instances:
+  - build:
+      context: .
+      env:
+        STRIPE_API_ENDPOINT: https://api.stripe.com/
+```
+
 ## Open Questions
 ### Build Inputs vs Configuration
 - Is there a difference between Docker build inputs and connector configuration files (currently the RFC does not distinguish these)? If so:
@@ -151,12 +197,10 @@ For example, for the NodeJS Lambda Connector, it could set the watch shell comma
   - How is the difference represented on disk in the user's Hasura project? (Different directories?)
 
 ### Custom CLI Plugins
-- Do connectors need to declare if they have a custom CLI plugin?
-- If custom CLI plugins provide configuration update services (such as DB schema introspection), does this need to be integrated with watch mode, and if so, how is this declared?
-- Do we need a `validate` subcommand to support the LSP/CLI?
+- `watchSubcommand` seems redundant if we're going to have a CLI contract. We could just have `watch: boolean` instead. Or perhaps the CLI plugin can declare what it supports and this exists outside of the connector-metadata.json?
 
 ### Publishing the Hasura Hub Connector Definition
-- How is this published to the Hub?
+- How is this published to the Hub? Git? `tar.gz`?
   
 ### OpenTelemetry
 - Do we want to reserve environment variables `OTEL_*` for possible future use of the [OTLP exporter spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md)?
--- a/rfcs/0001-packaging.md
+++ b/rfcs/0001-packaging.md
@@ -76,7 +76,7 @@ The `connector-metadata.json` contains JSON that describes:
 - The environment variables the connector supports to configure it (`supportedEnvironmentVariables`)
 - The packaging definition, which can be either `PrebuiltDockerImagePackaging` (a connector that does not require a build step), or  `ManagedDockerBuildPackaging` (a connector that requires a build step).
   - `PrebuiltDockerImagePackaging` defines the prebuilt `dockerImage` used to run the connector (`dockerImage`)
-  - If `ManagedDockerBuildPackaging` is used, a Dockerfile must be in the `.hasura` directory (and optionally, a `.dockerignore`). It will be used to build the connector.
+  - If `ManagedDockerBuildPackaging` is used, a Dockerfile must be in the `.hasura3` directory (and optionally, a `.dockerignore`). It will be used to build the connector.
 - A `commands` structure that optionally defines what shell commands to run for an "update" (eg. refresh schema introspection details) and "watch" (eh. watch and refresh schema introspection details periodically) scenario.
 - An optional `CLIPluginDefinition` that describes where to acquire the CLI plugin for this connector that can be used by the `commands` structure. If provided, the CLI plugin executable will be made available on the `PATH` for the commands and some configuration environment variables will be set (see the [CLI plugin RFC](https://github.com/hasura/ndc-hub/blob/cli-guidelines/rfcs/0002-cli-guidelines.md) for more details).
 
@@ -131,13 +131,13 @@ An example use case for this would be for a Postgres connector, where you may wa
 Note that the exact definition of how this needs to be specified in the Connector Definition will need to wait until how CLI plugins work is specified.
 
 ### Connector Layout in Hasura Project
-When a new connector is added to a Hasura project using `hasura3 add ConnectorManifest --hub hasura/nodejs-lambda:1.0`, the CLI acquires the Connector Definition for the specified Hub Connector. It then places this inside the `~/.hasura/hub-connectors/` directory. This is done to keep non-user editable files out of the user's source tree.
+When a new connector is added to a Hasura project using `hasura3 add ConnectorManifest --hub hasura/nodejs-lambda:1.0`, the CLI acquires the Connector Definition for the specified Hub Connector. It then places this inside the `~/.hasura3/hub-connectors/` directory. This is done to keep non-user editable files out of the user's source tree.
 
 The CLI then puts the `build-files/` into the build directory for that connector and adds a `.build.hml` file with the `ConnectorManifest` metadata object.
 
 ```
 # A copy of the `Connector Definition` for a particular hub connector version
-~/.hasura/hub-connectors/hasura/nodejs-lambda/1.0/
+~/.hasura3/hub-connectors/hasura/nodejs-lambda/1.0/
   connector-metadata.json
   docker-compose.yaml
   build-files/
@@ -173,7 +173,7 @@ definition:
   type: local
   connector:
     type: hub
-    name: hasura/nodejs-lambda:1.0 # This is used to find the Hasura Hub Connector Definition in ~/.hasura/hub-connectors/
+    name: hasura/nodejs-lambda:1.0 # This is used to find the Hasura Hub Connector Definition in ~/.hasura3/hub-connectors/
   tunnel: true
   instances:
   - build:
--- a/rfcs/0001-packaging.md
+++ b/rfcs/0001-packaging.md
@@ -78,11 +78,11 @@ The connector definition package is a .tgz (tar file, gzipped) that contains the
   .hasura/
     connector-metadata.json # See ConnectorMetadataDefinition type
     docker-compose.yaml # If DockerComposeWatchMode is used
-  
-  ### Start: If Docker packaging requires a build step
-  .dockerignore
-  Dockerfile
-  ### End: If Docker packaging requires a build step
+    
+    ### Start: If Docker packaging requires a build step
+    .dockerignore
+    Dockerfile
+    ### End: If Docker packaging requires a build step
     
   ### Start: Connector specific configuration files
   src/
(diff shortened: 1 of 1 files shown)


rfcs/0001-packaging.md
```diff
@@ -0,0 +1,46 @@
...
+- The connector executable should accept the following subcommands:
+  - `serve` should start a HTTP server on port `8080`, which is compatible with the NDC specification, with `/` as its base URL.
+    - For example, `http://connector:8080/query` should implement the query endpoint
+    - The default port can be overwritten using the `HASURA_CONNECTOR_PORT` environment variable.
+- The image entrypoint should be set to the connector process, and the default `CMD` should be set to the `serve` command.
+- The connector can read environment variables on startup for configuration purposes
+  - The following environment variables are reserved, and should not be used for connector-specific configuration:
+    - `HASURA_*`
+  - Connectors can use environment variables as part of their configuration. Configuration that varies between different environments or regions (like connection strings) should be configurable via environment variables. 
+- The connector should send any relevant trace spans in the OTLP format to the OTEL collector hosted at the URL provided by the `HASURA_OTLP_ENDPOINT` environment variable.
```
BenoitRanque, line 31: Continuing from my comment on the doc:
The [OTEL spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#configuration-options) indicate that the env var `OTEL_EXPORTER_OTLP_ENDPOINT` (and 3 others) must be used by exporters to specify the collector endpoint.

Accordingly, OTEL [libraries](https://github.com/open-telemetry/opentelemetry-rust/blob/8237f87f42aa6d4935ed6a7e9fb9b225fb20640e/opentelemetry-otlp/src/exporter/http/mod.rs#L304) read that env var. At least that's my current understanding.

If we do define our own `HASURA_OTLP_ENDPOINT` env var, it's my understanding that connector implementers will need to read that env var, and feed the value into their OTLP exporter.

Do note, I'm new to OTEL, so I could be missing something. I'm hoping that using `OTEL_EXPORTER_OTLP_ENDPOINT` would allow at least some OTEL SDKs to work out of the box.

daniel-chambers:
I've made some draft updates to this spec after our meeting about how to handle connectors that have build steps and how to handle connectors that require special watch mode configuration. I've put them on a [separate branch](https://github.com/hasura/ndc-hub/blob/rfc/packaging-spec-daniel/rfcs/0001-packaging.md), but I can merge that to this PR if we're OK with the changes.

scriptnull:
A small note here: I might write a "distribution spec" for packages adhering to this packaging spec.

Initial thoughts: We might want a tool that can produce a file called `package.tgz` for every new version of a connector. That file will then be published to our public connector hub registry (which is a Google cloud bucket to start with). That file will then be downloaded by the CLI to figure out how to run the connector locally.

(I'm going to log off for the day right now, but if anyone is interested in hearing more around it please let me know)

daniel-chambers:
Should the hub packaging and manifest be specified in this RFC? Currently this talks specifically about how to package the connector itself, but no _specifics_ about how you (as a connector author) deliver that to the Hasura Connector Hub and how you describe yourself (watch config, packaging type, version, description, etc etc) to the Hub, for consumption by Hasura CLI etc tooling.

I tend to think this RFC should discuss those things, but WIP details about that are currently in our DX google doc. I feel like, to make this spec more complete, it really needs to discuss those details as well, which would then cover what @scriptnull mentions above.

cc @paf31 @nullxone

paf31:
@daniel-chambers Yes, I think we should add those details here too. I was waiting for some input from the CLI team, but I think we can take a first stab at it. Or alternatively, we can merge this now, if everyone is happy, and make a follow-up RFC for the hub manifest spec.

daniel-chambers:
@paf31 I have added some information about the Hub Connector Definition to the RFC.

rfcs/0001-packaging.md
```diff
@@ -0,0 +1,198 @@
...
+```
+/
+  .hasura/
+    connector-metadata.json # See ConnectorMetadataDefinition type
+    docker-compose.yaml # If DockerComposeWatchMode is used
+  
+  ### Start: If Docker packaging requires a build step
+  .dockerignore
+  Dockerfile
+  ### End: If Docker packaging requires a build step
```
daniel-chambers, line 85: We could put these files in `.hasura` and the tooling could just invoke `docker build -f ./hasura/Dockerfile`, if we wanted to keep them out of the way.
I left them in the directory, because the Dockerfile makes more sense when you read it when it is located in the same directory as its build context is. I'd not wedded to that if we want to keep tooling related stuff tucked away inside `.hasura` though.

nullxone: Yep, let's tuck the implementation details away inside the .hasuraconnector folder.

nullxone:
Hey, looks good to me! @paf31 @daniel-chambers 

I have already reviewed this RFC and provided a response [here on the Slack thread](https://hasurahq.slack.com/archives/C05NBS2MYHM/p1706814855427819?thread_ts=1706608624.284649&cid=C05NBS2MYHM)

@paf31 Is there anything else specific that the CLI team or I can pitch in and help with?

As far as I can tell, the current DX spec at following documents remain unchanged except for the addition of docker compose based watch mode. The following documents also contain the interface for Hasura Hub definitions.
* [Hasura v3 DX Implementation](https://docs.google.com/document/d/1tgMW5OnNuZrothsoIeo9t8ZI0WZKKo5s3HUMVbuWBbM/edit?pli=1#heading=h.n8cw8tt9qy7)
* [Connector Developer <> CLI Developer Gist](https://gist.github.com/nullxone/a0c278bba49874fb0a280369faca3ebd)

I would suggest that we move to implementation for the typescript and postgres connectors and allow the remaining details to be filled in by discovery of the best possible implementation for each case.

@manasag @ManushiKhanna

nullxone:
Hi team, I took a crack at the Hasura Hub manifest, and merged this spec into [this gist](https://gist.github.com/nullxone/a0c278bba49874fb0a280369faca3ebd)

Key changes:
* Always use docker compose watch. Remove other watch modes for now.
* Move watch mode config one level up, because we can also watch docker image based connectors in the same way (I think; or can be made equivalent)
* Remove docker-compose file, add a watch key into the configuration json. This ensures consistent behaviour between local and cloud. CLI to generate the docker compose file and run it.
* Add plugin information such as downloadUrl and scripts interface for CLI.

@paf31 @daniel-chambers let me know what you think.

nullxone:
For completeness, there is also
* additional information as currently being tracked by Hasura Hub. [Example here.](https://github.com/hasura/ndc-hub/blob/main/registry/typescript-deno/metadata.json)
* [Phil's packaging spec on Google Docs](https://docs.google.com/document/d/1R20TLVoFMlXAvSdGtqHADUFdk6BIHKMj-qeJZmOELOM/edit?pli=1#heading=h.4vemujta577m): adds information about documentation and branding, and contains notes on multitenant implementation

I think we can get to this stuff in a non blocking fashion. wdyt?

rfcs/0001-packaging.md
```diff
@@ -0,0 +1,198 @@
...
+  type: "PrebuiltDockerImage"
+  dockerImage: string // eg "hasura/postgres-data-connector:1.0.0"
+}
+
+type DockerBuildPackaging = {
+  type: "DockerBuild"
+  watchMode: WatchMode
+}
+
+type WatchMode = DockerRebuildWatchMode | DockerComposeWatchMode | ShellCommandWatchMode
```
paf31, line 115: Do we want to just use the existence of a docker compose config to indicate this? Same for shell command watch mode, I guess, which could be put in an executable file under `.hasura` with a specific name. If the user specifies `DockerComposeWatchMode` but then deletes the config file, we'd have to error, right? But if we use the files as the source of truth, they can just delete or rename it to use the default watch mode.

rfcs/0001-packaging.md
```diff
@@ -0,0 +1,198 @@
...
+### Hasura Hub Connector Definition
+In order for the Hasura tooling to understand a connector, know how to interact with it and know how it is packaged, a connector will need a definition that contains this information.
+
+The connector definition package is a .tgz (tar file, gzipped) that contains the following files:
+
+```
+/
+  .hasura/
+    connector-metadata.json # See ConnectorMetadataDefinition type
+    docker-compose.yaml # If DockerComposeWatchMode is used
```
paf31, line 80: We only need the watch section of this, right? Because we don't want the user to think they can define multiple containers.

nullxone: Can we roll the docker compose watch into the configuration.json?

This ensures consistent behaviour between local and cloud. CLI to generate the docker compose file and run it.

Here's example types that support this:
https://gist.github.com/nullxone/a0c278bba49874fb0a280369faca3ebd

nullxone: Basically, the watch key is the ONLY thing they can define. Not build, args, context, volumes, or anything else. Because all the rest is standardized, and should be generated by the CLI for consistency.

daniel-chambers: There are a couple of reasons why I used the whole docker compose file:
- Save the tooling the bother of generating one, just use what is defined by the connector author
- No need for importing Docker Compose watch config types into our JSON schema, and then the bother of keeping that up to date if/when Docker change their types
- Let the container author do (almost) whatever they want in there, so long as they expose the necessary port, and document our recommendations/requirements.

Yeah, this is the opposite approach to locking it all down to precisely what we want to allow and no more, but it feels a bit more open and potentially flexible to use cases we haven't considered.

rfcs/0001-packaging.md
daniel-chambers: What do you do if they put both files there (a compose file and a shell file)? I guess I prefer explicit intentions to be expressed in one spot and every other invalid state becomes errors.

I'm not wedded to this position, however, but I usually prefer being explicit than being implicit.

daniel-chambers:
> * Always use docker compose watch. Remove other watch modes for now.
I'd like to see the Shell Script watch mode in there asap, it is useful for debugging NodeJS functions.
 
> * Move watch mode config one level up, because we can also watch docker image based connectors in the same way (I think; or can be made equivalent)
I don't think this is true. How does one "rebuild" a container image (a feature of docker compose watch) without a Dockerfile (ie. image-only)? This is why I kept it for DockerBuild only.

> * Remove docker-compose file, add a watch key into the configuration json. This ensures consistent behaviour between local and cloud. CLI to generate the docker compose file and run it.
I put my thoughts about why I prefer having an explict docker-compose file in a previous comment.

> * Add plugin information such as downloadUrl and scripts interface for CLI.
👍

rfcs/0001-packaging.md
```diff
@@ -0,0 +1,198 @@
...
+
+- The result of the Dockerfile build should be a connector Docker image that contains the configuration files and can start immediately.
+
+### Common Docker Image Behaviour
+The Docker images of both categories of connectors behave in the same way with respect to the following points:
+
+- The connector executable should accept the following subcommands:
+  - `serve` should start a HTTP server on port `8080`, which is compatible with the NDC specification, with `/` as its base URL.
+    - For example, `http://connector:8080/query` should implement the query endpoint
+    - The default port can be overwritten using the `HASURA_CONNECTOR_PORT` environment variable.
```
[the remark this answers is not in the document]
daniel-chambers: I think this is reasonable. It is an extremely standard env var.

rfcs/0001-packaging.md
daniel-chambers: Done, the files have been moved into the `.hasura` directory.

rfcs/0001-packaging.md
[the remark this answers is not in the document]
daniel-chambers: Done.

daniel-chambers:
> - Add plugin information such as downloadUrl and scripts interface for CLI.

Actually, looking at this again, I don't understand what's going on in your Gist suggestion. What's this `scripts` thing? What's a "sidecar" (sidecar usually refers to a secondary Docker container that provides some middleware service to a primary container; we don't have that concept here)? I don't know why there's a primary and a sidecar?

daniel-chambers:
I've updated the [Open Questions](https://github.com/hasura/ndc-hub/blob/4f0ce1920518382889cf6ec216cbbc8462609d0f/rfcs/0001-packaging.md#open-questions) part of the RFC and added some new ones. 

Please comment as to your thoughts.

rfcs/0001-packaging.md
```diff
@@ -0,0 +1,213 @@
...
+
+## Open Questions
+### Build Inputs vs Configuration
+- Is there a difference between Docker build inputs and connector configuration files (currently the RFC does not distinguish these)? If so:
+  - Can connector configuration be optional (defaulted to zero files) where it is not used (ie NodeJS Lambda Connector)
+  - How is the difference represented on disk in the user's Hasura project? (Different directories?)
+
+### Custom CLI Plugins
+- Do connectors need to declare if they have a custom CLI plugin?
+- If custom CLI plugins provide configuration update services (such as DB schema introspection), does this need to be integrated with watch mode, and if so, how is this declared?
```
daniel-chambers, line 206: My suggestion is to update the manifest types like so:

```typescript
type ConnectorMetadataDefinition = {
  packagingDefinition: PackagingDefinition
  supportedEnvironmentVariables: EnvironmentVariableDefinition[]
  cliPlugin?: CliPluginDefinition
}

type CliPluginDefinition = {
  pluginDownloadUrl: string // https://url/to/download/plugin/to/install.tgz
  pluginPackageSHA256: string // To ensure plugin integrity and also to identify which version to use out of those installed (I can see us needing side-by-side versioning to support immutable builds)
  watchSubcommand?: string // If the connector wants to add to the watch mode, it can define what subcommand to use "watch --whatever --flags"
}
```

sordina:
Hi All,

Just adding some feedback/questions.

* Git mechanism for Connector Hub connector configuration and templating
* Archive mechanism for build service
* Watch mechanisms

### Connector Hub Connector Configuration and Templating

There seems to be a conflation of the tar/archive for build purposes (makes sense) and for connector authorship purposes (skeptical). In my opinion, levering the source control flows for connector authorship should be the default process for several reasons:

1. Transparency - This avoids blobs that can't be quickly inspected with the usual online and local tools
2. Changes - Blobs can't be diffed, merged, etc.
3. Processes - There are established project management flows such as PRs etc that can be taken advantage of in normal repos with normal directories and files
4. APIs for archive fetch - The end user when leveraging hub connectors doesn't need to use source control directly, this can be behind APIs that expose the tar/archive interface described. This is primarily a benefit for authors and reviewers

### Archive Mechanism for Build Service

This is about bundling config and resources to send to the metadata build service. Using an archive mechanism here makes sense, although the argument could be made that a source-control/differential approach could be more efficient in some circumstances. I think starting with tar is the right approach.

### Watch Mechanisms

I had some questions about the final form of the watch mechanisms.

There seem to be several potential compoents all interacing:

* Hub config and metadata
* local watch command config
* connector plugin config
* connector plugin hard coded watch behaviour
* normal docker-compose files (shared)
* docker-compose snippets - inlined, not shared

Now my understanding is that the actual design will be as follows:

* Default project configuration for a connector will be stored on the hub and downloaded when creating a new connector
* A docker-based watch command will be used by default leveraging the docker-compose snippets approach, or
* If the connector has a plugin installed the plugin based watch command will be used - This is free to watch in whatever manner it chooses

Is this a correct understanding?

If so I think I'm happy with this.

daniel-chambers:
## Meeting notes

* There are two top-level types of Data Connectors
  * Non-Hasura Hub Connector built from a Dockerfile in the user's Hasura project
  * Hasura Hub Connector - built consistently from a definition defined in the Hasura Hub
* Hasura Hub Connectors have two types
  * Pre-built Docker Image
  * Managed Docker Build (connector provides a Dockerfile used to build, and docker-compose.yaml to watch - not customisable by the user)
* Managed Docker Builds _could_ be implemented in terms of a Non-Hasura Hub Connector (ie. build from a Dockerfile). 
  * However, this could create support problems by allowing the user to customise too much and break things, which support then has to diagnose from first principles (ie check they haven't broken the Docker plumbing accidentally)
  * Managed Docker Builds also allow us to encapsulate how this is done and potentially change it in the future
* `.hasura` folder is a cache of Hasura Hub Connector definition information for tooling use
  * Not supposed to be edited by the user. How can we prevent them from doing so?:
    * Documentation
    * We could put comments in the files saying don't touch
    * Rename the files and add `.lock` to try to imply they are like lockfiles (ie. not user editable)
    * Don't put the `.hasura` folder in the user's working tree, put it in a cache dir in their profile
* The connector manifest needs a way of specifying a CLI plugin exists and then using it to perform a custom connector "watch"
  * Is not the same "watch" as watching the connector configuration files and restarting/rebuilding the container. Is an addition to that.
  * Use case: Watching the Postgres DB schema and updating schema introspection configuration file
  * Issue raised where the user may want different watch workflows. Example: some users may not want their DB schema watched because it is a brownfield rarely changing schema. Options:
    * Make the user run the watch manually and therefore choose
    * Run the watch automatically as a part of `hasura dev` (smooth DX, works out of the box) and give them a way to customize the 
behaviour. Could be a config file that the CLI plugin looks at to determine whether to poll the DB or not.
  * The CLI plugin needs a standard interface that allows the tooling to pass env vars and the context directory when it is invoked via `hasura3 dev`
* Instead of connector authors tarring up their Hub Connector Definition and submitting it to a service, why not get them to link a git repo with the files in it, and track tags/branches (TDB)? If tooling needs tarfiles to download the information, the Hub service can do that.
* docker-compose.yml can be defined for both PrebuiltDockerImage and DockerBuild to define how to watch.
  * In the Prebuilt case, the compose file should not define a build property or an image property, and the tooling will import it and set the image itself to what is defined in the connector-metadata.json (could we use a variable for this instead, like PORT?)
  * This unifies how the CLI watches for both cases (ie via docker compose)
* ShellCommandWatch mode is intended as an escape hatch to support the use case of running local tooling (eg using `npm run watch` for NodeJS Lambda Connector for the purposes of attaching a debugger to the local node process)
  * We will want multiple watch strategies available and a way for the user to customise the default
* No, we don't need a validate subcommand to support the LSP/CLI. Running the connector and it failing to start with an error is what does the validation.

Another meeting has been scheduled for tomorrow to continue the discussion. I will try to update the RFC tomorrow with changes to support the above discussion points.

rfcs/0001-packaging.md
```diff
@@ -0,0 +1,213 @@
...
+
+### Common Docker Image Behaviour
+The Docker images of both categories of connectors behave in the same way with respect to the following points:
+
+- The connector executable should accept the following subcommands:
+  - `serve` should start a HTTP server on port `8080`, which is compatible with the NDC specification, with `/` as its base URL.
+    - For example, `http://connector:8080/query` should implement the query endpoint
+    - The default port can be overwritten using the `PORT` environment variable.
+
+- The image `ENTRYPOINT` should be set to the connector process, and the default `CMD` should be set to the `serve` command. This can mean setting it to the connector executable itself, or some intermediate executable/script that eventually provides its command line arguments to the connector executable.
```
SamirTalwar, line 50: Why do we need to specify that the `serve` subcommand is required? If we're expect that the Docker image should specify the `CMD`, then just let the image dictate if a subcommand is required.

daniel-chambers: Perhaps this made more sense when we had multiple possible subcommands because of the configuration server. However, this has been moved by @paf31 to https://github.com/hasura/ndc-hub/pull/94.

daniel-chambers:
## Round 2 Meeting Notes
* Decided that we will hide the difference between config files and Docker build inputs from the end-user. 
  * For `PrebuiltDockerImage` connectors, what is found in the build directory in the user Hasura project is the config files. (`PrebuiltDockerImage` connectors don't have build inputs by definition)
  * For `ManagedDockerBuild` connectors, what is found in the build directory in the user Hasura project is the build input files. When this connector is submitted to the build service, an empty archive of config files will be submitted for it. The connector will not use the volume mounted to the `/etc/connector` folder and will use whatever it builds into its image.
  * This approach ensures that both types of connector packaging are run the same (both have a volume mounted, just one doesn't care about what's in it)
  * We decided that the end-user doesn't really care about the distinction between config files and build inputs so we'd rather hide it in the DX for now
  * This does mean (for now) we don't support the use case of a connector that uses both a build input _and_ config files
* We agreed that for Hasura Hub Connectors, their Connector Definition will be kept/cached out of the users' source tree in the `~/.hasura` directory somewhere
  * However, we still want to support the use case of a non-Hasura Hub Connector where you can just copy the necessary files (ie the Connector Definition) into the Hasura project source tree and it just works. The RFC needs to be updated to add and document this capability.
* We decided to cut `ShellCommandWatchMode` (and `additionalWatchModes`) from scope. The scenario it addresses (running a connector locally using local non-Docker tools) can looked at a later date
* We decided that the how the Connector Definition is published to the Hasura Hub can be left to another RFC. We can, for now, just assume that a tar.gz of the Connector Definition can be published somewhere.
* We discussed the CLI plugin and decided that the details of how that's being implemented (Docker vs native) can be left outside of this and what's in the Connector Definition can just be a placeholder
  * We did say that the `watchCommand` should continue to define the command passed to the plugin to start the plugin's watch mode and that it's okay that the tooling makes some assumptions about extra parameters it will append to that command to pass information like env vars and context dir.
* The open question about OpenTelemetry env vars has been moved to the deployment spec and answered there (yes, we will reserve that env var namespace)

I will update the RFC document with the necessary changes tomorrow.

daniel-chambers:
I have updated the RFC document with the changes discussed in the Round 2 meeting.

@nullxone I've modified the `ConnectorManifest` to be able to describe the difference between a connector from the Hub vs an "inlined" connector (ie imported into the project from a non-Hub source. ie a tgz file or wherever). Please take a look and let me know your feedback.

rfcs/0001-packaging.md
```diff
@@ -0,0 +1,232 @@
...
+#### Non-Hasura Hub Connector
+To support the use case of a connector developer being and to share/iterate on a connector and use it inside a Hasura project without first publishing it to the Hasura Connector Hub, we will support the ability to inline a Connector Definition inside the Hasura project source tree. This enables the Hasura tooling to still know how to interact with the connector, but does not require it to be published on the Connector Hub.
+
+The CLI could support the use of non-Hasura Hub Connectors by supporting the use of `.tgz` archives of the Connector Definition: `hasura3 add ConnectorManifest --inline my-connector-definition.tgz`
+
+```
+project/
+  subgraphs/default/dataconnectors/
+    my-dataconnector/
+      connector-definition/ # `my-connector-definition.tgz` extracted into the source tree
```
paf31, line 191: It seems unnecessary to have two copies of `build-files` here. If the `project/{connector-dir}` directory was considered an overlay on top of `~/.hasura/hub-connectors/{connector-dir}` (when that directory exists), then that accommodates all use cases here, plus some more (e.g. the user _could_ modify `Dockerfile` for the TS connector, but it just wouldn't be copied into `project/` by default). In the inline case, there is nothing to overlay on top of, so `project/` would contain its own `build-files`, `docker-compose.yaml`, etc.

rfcs/0001-packaging.md
```diff
@@ -0,0 +1,232 @@
...
+
+Connector behavior is specified by the [NDC specification](http://hasura.github.io/ndc-spec/), and the deployment API is specified in the [deployment RFC](./0000-deployment.md). This document specifies how the Hasura CLI and Hasura Connector Hub can work together to facilitate connector development, packaging and deployment via the API.
+
+## Proposal
+
+### Docker Packaging Categories
+
+Hasura Hub data connectors are packaged as Docker images which follow the [deployment specification](./0000-deployment.md). Per the deployment specification, there are two ways in which the CLI can describe those images to the deployment service:
+
+- Connectors that do not require a build step can be described using a name and version, which should correspond to an entry in the connector hub registry.
```
scriptnull, line 13: Hi, we were trying to work out the "local" and "cloud" user flow of how running these types of connectors would look. We have some doubts regarding it.

1. A user is trying to use the [ndc-clickhouse](https://github.com/hasura/ndc-clickhouse) connector in their project. This connector does not have a docker build step.
2. So, they do `hasura3 add connectormanifest billingdb --hub hasura/ndc-clickhouse:v0.1.0` 
3. CLI downloads the tar ball package that satisfies this spec and unpacks the tar ball for using.
4. CLI looks at `connector-metadata.json` file which might be like this
  ```js
 {
   "packagingDefinition": {
      "type": "PrebuiltDockerImage"
      "dockerImage": "hasura/ndc-clickhouse:v0.1.0"
    },
   "supportedEnvironmentVariables": [
     {
        "name": "username",
        "description": "user name to access click house DB"
      },
     {
        "name": "password",
        "description": "password to access clickhouse DB"
     },
     {
        "name": "url",
        "description": "url to access clickhouse DB"
      }
    ],
    "cliPlugin": {
        // fill in the blanks
    }
 }
  ```
5. By looking at the above file, I think CLI would try to construct and write the ConnectorManifest object in the user's local cli project.
  ```yaml
  kind: ConnectorManifest
  definition:
    name: cman1
    type: local
    hub-connector: hasura/clickhouse:v0.1.0

    # has a map of supportedEnvironmentVariables here
  ```
6. Since the type is `local`, CLI will try to run the docker container and set the environment variables present in the manifest to the docker container that was created from `hasura/clickhouse:v0.1.0` image. 
7. But if we take a look at https://github.com/hasura/ndc-clickhouse, the clikchouse connector is expecting for a `confiugration.json` file of the following format to work.
  ```
  {
     "connection": {
       "username": "your_username",
       "password": "your_password",
       "url": "your_clickhouse_url"
     },
     "tables": [
       {
         "name": "TableName",
         "schema": "SchemaName",
         "alias": "TableAlias",
         "primary_key": { "name": "TableId", "columns": ["TableId"] },
         "columns": [
           { "name": "TableId", "alias": "TableId", "data_type": "Int32" },
         ]
       }
     ]
   }
  ```
8. Can clickhouse connector understand the env vars sent by CLI and use them for values in `connection` field? or should the CLI be taking care of constructing this `configuration.json` file and mount it to clickhouse container?

scriptnull: A small request: Could we prepare the actual tarball based on this spec for the clickhouse connector alone to start with? That way we could use it and verify the end to end flow for it.

daniel-chambers: @scriptnull This doesn't seem like an issue with this spec, this seems like the issue where we've made all these breaking changes to how things are supposed to work and not everything has been updated to support the various new specs and changes. 😅

If the Clickhouse connector is still taking a configuration.json with the connection string etc, then it hasn't been updated to support the new deployment spec. This isn't surprising, the PR to update the Rust SDK to support the [new deployment spec changes](https://github.com/hasura/ndc-hub/pull/97) hasn't merged yet.

When the Clickhouse connector has been updated, I'd expect it to take those connection details as env vars and not in a configuration file.

Attn: @BenoitRanque

rfcs/0001-packaging.md
```diff
@@ -0,0 +1,232 @@
...
+  - If `ManagedDockerBuildPackaging` is used, a Dockerfile must be in the `.hasura` directory (and optionally, a `.dockerignore`). It will be used to build the connector.
+- An optional `CLIPluginDefinition` that describes where to acquire the CLI plugin for this connector that can be used to enhance watch mode with connector-specific configuration updates
+
+### Watch Mode
+When developing locally using the connector, the user may utilize the connector in a "watch mode" that automatically reloads the connector with new configuration as they change the configuration.
+
+Every connector must specify a `docker-compose.yaml` in their connector definition that defines how to watch the container.
+
+#### Connectors that do not require a build step (`PrebuiltDockerImagePackaging`)
+The `docker-compose.yaml` should contain something like this:
```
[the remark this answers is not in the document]
daniel-chambers: @nullxone wanted to always have it there for consistency purposes. I agree that technically the CLI tooling could generate it entirely for `PrebuildDockerImagePackaging`.

rfcs/0001-packaging.md
daniel-chambers: It is true that this directory in this location is useless in the case of the Non-Hasura Hub Connector (the `build-files` directory does not work in an "overlay" fashion; it is a starting template). However, keep in mind when the CLI uses the `.tgz` file to create everything, it would use the `build-files` directory in the `.tgz` to create the `build` directory, so the directory itself isn't useless. 

The `connector-definition` folder is just the `.tgz` extracted out. Technically the CLI could delete the `build-files` directory after extraction (and after copying it to the `build` directory), for cleanliness. @nullxone, do you have any thoughts on this?

rfcs/0001-packaging.md
imperfect-fourth, line 13: Are the "connectors that do not require a build step" expected to not have ANY configuration files or context volumes?

daniel-chambers: No, this is just referring to how to build a connector (mostly described in that linked deployment spec). Once a connector is "built" and ready to deploy, they always get their config files mounted to `/etc/connector` (see that other spec).

daniel-chambers:
I just had a meeting with @SandeepSamba and we hashed out how we wanted the CLI plugins to be represented in the connector manifest. (See [3ea1ee1](https://github.com/hasura/ndc-hub/pull/89/commits/3ea1ee157d0097ccc1fb2d2397d0782615957b2e))

There we a couple of issues that remain unsolved that probably live outside of this RFC but nonetheless need considering:
* Port mapping - if a user users multiple connectors in a Hasura project, and they all serve on 8080 by default, what performs the port mapping to give each connector a unique port? The CLI? The user, in some configuration? The CLI automatically determining free ports and allocating them would be the best user experience, but more difficult to develop. A more manual approach may be to add a port setting to the ConnectorManifest (type: local) to allow the user to define the local port.
* Command overrides - @SandeepSamba raised the issue that the user may want to override the default commands provided by the connector for `update` and `watch`. One suggestion was to add a `commandOverrides` to the ConnectorManifest, where the user could specify alternative commands to use.

@nullxone do you have any comment on the above points? The port issue is more pressing, command overrides can be figured out later IMHO.

nullxone:
1. Regarding structure of the tarfile. If there are no good reasons otherwise, could we please keep it the same as the directory structure of connectors in a project? This way it is easier to develop and distribute connectors. Importantly, this can be mapped to the alternative directory structure proposed in this spec, making it equivalent.
```
tarfile.tgz/
  .hasuraconnector/
    connector-metadata.json
    docker-compose.yaml
    # other files to define build and run
  tables.yaml
  functions.ts
  package.json
  # other user editable connector config files
```

2. Regarding port mapping. CLI can figure out free ports. A cursory search yielded the following resources. Should be reasonable.
https://stackoverflow.com/questions/8817500/how-to-make-sure-that-a-certain-port-is-not-occupied-by-any-other-process
https://stackoverflow.com/questions/28989069/how-to-find-a-free-tcp-port

We should also allow explicit configuration in ConnectorManifest, but we can add this later.

3. Something like commandOverrides is an excellent idea. I am tracking many good change ideas to the ConnectorManifest: ports, how to specify type local/hub (connector type syntax from Daniel on slack thread), more comprehensive structure changes discussed with Vishnu and Divi. I also expect additional changes once the build system is finalised.

In the interest of presenting a stable spec to the dx team, I will continue to collect these ideas for now. We can design and iterate on this post an initial release, and apply all these changes as a single changeset.

@daniel-chambers

rfcs/0001-packaging.md
imperfect-fourth: FYI @scriptnull :point_up_2:

nullxone:
Hey @daniel-chambers , completely on the same page about all that stuff, about the possibility of hiding away the .hasuraconnector folder (build and run instructions) into a home directory. Having said that,

1. This requirement is independent of the packaging structure. The CLI is ultimately responsible for the project directory structure, and as long as we have _some_ equivalent packaging structure, the CLI can arrange it in the correct way in the end.

2. We had discussed wanting to support two configurations. One where the .hasuraconnector folder is hidden away in a user's home directory, and another where it was inlined into the project folder structure for custom connectors or user edits. All I meant is that this is an opportunity where we can simplify the packaging structure to reflect the project structure in one of these cases, meaning that it becomes trivial for the connector developer to package their custom connector if they want to, they just zip up that directory. CLI can move out the .hasuraconnector folder _anyway_.

Intended to respond to your comment above, is there a more specific sense in which I can comment on this?

> The connector-definition folder is just the .tgz extracted out. Technically the CLI could delete the build-files directory after extraction (and after copying it to the build directory), for cleanliness. @nullxone, do you have any thoughts on this?

Also don't intend to cause churn to the spec, packaging structures both seems equivalent to me, because ultimately the CLI can arrange these folders somehow (packaging structure can be independent from project structure, but need not be!). Your call on if this makes sense.

nullxone:
@daniel-chambers to add to point 2 above, as we'd previously discussed, we may also want to inline the .hasuraconnector folder for cases where the connector is distributed through channels other than hasura hub so that a stable connector definition can be checked in to the project repository. For example, connectors directly fetched from github or tarfile urls.

nullxone:
@daniel-chambers To state this another way, just because a connector was distributed through github and not hasura hub, does not mean that the end user needs to interact with the connector developer files. They should still remain separate concerns.

daniel-chambers:
@nullxone I agree with you in principle on your points; we definitely want the ability for connector developers to zip up their connector folder and share it outside the hub (although I do think it is worth pointing out that this should be a minority use case, and publishing to the hub should be preferred and encouraged, as it simpler to find, consume and manage for end users).

I believe the existing RFC structure already gives us the ability to zip up the connector directory and redistribute it. Here's the example directory from the "Non-Hasura Hub Connector" section with some bits removed for clarity:

```
project/
  subgraphs/default/dataconnectors/
    my-dataconnector/
      connector-definition/
        connector-metadata.json
        docker-compose.yaml
        docker/
          .dockerignore
          Dockerfile
      build/
        src/
          functions.ts
        package.json
        package-lock.json
        tsconfig.json

        my-dataconnector.build.hml
```

To redistribute this, the user could zip the `my-dataconnector` folder.

However, I don't think I've done a good job of explaining why I think it's better to make the connector definition file layout not match the build directory contents, which is what you're suggesting. Let me attempt to do a better job.

To compare, this is what the folder structure would look like if we decided to make the connector definition file layout match the contents of the project build directory, as in your suggestion (contents of `connector-definition` folder moved into `build/.hasuraconnector` and flattened:

```
project/
  subgraphs/default/dataconnectors/
    my-dataconnector/
      build/
        .hasuraconnector/
          connector-metadata.json
          docker-compose.yaml
          .dockerignore
          Dockerfile
        src/
          functions.ts
        package.json
        package-lock.json
        tsconfig.json

        my-dataconnector.build.hml
```

While this would work, my issue is that this is placing the connector definition in the build context for the Dockerfile. The CLI would need to add `.hasuraconnector` to the `.dockerignore` file at runtime to ensure the directory is not included in the build context (which would leave the file dirtied in the user's working copy, unless reverted). We could insist the connector developer manually put `.hasuraconnector` in the `.dockerignore`, but then it doesn't make sense when the connector is published as a Hub Connector and the `.hasuraconnector` folder isn't in the build directory and doesn't need to be ignored. By keeping the connector definition separate from the build directory, we ensure the build behaviour is identical for when the connector definition is inlined vs. when it is in the `~/.hasura3` directory.

At this point, in the interests of just doing something and moving on, if you still feel strongly that you prefer the `.hasuraconnector` layout, I'll update the RFC to just do that and we can merge it. Other than the build context concerns from above (which I do think are important from a consistency perspective), the difference is mostly aesthetics from my perspective and isn't worth holding this RFC up over any more. We need to get started building stuff on this asap.

0x777:
Regarding the layout, ideally I would've loved something like this:

```
build/
  .hasuraconnector/
  myconnector.build.hml
  config/
    tables.hml
    functions.ts
```

I feel that all three of these (the definition, ConnectorManifest and config) are siblings. If you look at myconnector.build.hml, it refers to a definition - either to one stored on Hasura hub (maybe cached in user's home dir) or one stored locally in a directory. The natural place for this directory seems next to the ConnectorManifest than in a parent folder.

The `config` directory has the config for a `PrebuiltDockerImagePackaging` or build context for a `ManagedDockerBuildPackaging` connector. When the connector author writes `.dockerignore` they are only worried about what they should typically ignore in that language/connector's environment (node_modules for eg. in nodejs projects).

I think Akshaya's point is that this introduces too much nesting and wants to flatten the `config` directory into `build`. However, once we do that, it introduces the .dockerignore problem that you described above. Between these two, I'm less concerned about the `.hasuraconnector` not being part of `.dockerignore`. We'll need to solve the `.dockerignore` problem anyway - the user may want to ignore files and folders other than just `.hasuraconnector`.

I don't care much about the use case of 'packaging the build directory as a new connector'. I'm not convinced that it is the right paradigm. For example, if I were to build a Sendgrid connector on top of a nodejs connector, my `*.js` files shouldn't be unpacked into the user's directory - they are not meant to be modified but should be included in the docker image for the connector.

To summarize, my recommendation would be to have the connector definition (.hasuraconnector) next to the file that defines ConnectorManifest (build.hml). This would also mean that build-files/ isn't part of `.hasuraconnector`.

rfcs/0001-packaging.md
```diff
@@ -0,0 +1,237 @@
...
+### Connector Definition
+In order for the Hasura tooling to understand a connector, know how to interact with it and know how it is packaged, a connector will need a definition that contains this information.
+
+The connector definition is a file structure that contains the following files and directories:
+
+```
+/
+  connector-metadata.json # See ConnectorMetadataDefinition type
+
+  docker-compose.yaml
```
0x777, line 25: We've decided that `docker-compose.yaml` would be removed and the `watch` section will more into `connector-metadata.json`.

daniel-chambers: Updated.

nullxone (approved):

sordina (approved):
Let's Go!

PR #91, opened 2024-02-01
Add packaging categories and watch mode information

paf31:


rfcs/0001-packaging.md
```diff
@@ -14,33 +14,84 @@ _This RFC does not specify the following planned changes:_
...
 
-## Open Questions
+### Watch Mode
+When developing locally using the connector, the user may utilize the connector in a "watch mode" that automatically reloads the connector with new configuration as they change the configuration.
+
+#### Connectors that do not require a build step
+If any configuration file changes, the connector image is restarted with the new version of the configuration file mounted.
+
+#### Connectors that require a build step
+These connectors can provide a [docker compose watch configuration](https://docs.docker.com/compose/compose-file/develop/#watch) that defines how to watch the configuration files and how to react to different files changing. It can react by causing a container rebuild and restart, or by copying the new file inside the existing container and optionally restarting the container. This allows the connector to optimise its hot reload capability by only performing a rebuild where necessary.
```
paf31, line 76: > These connectors can provide

Specifically, the connectors might provide a default config as part of the hub manifest, but the user is responsible for providing the watch config that's actually used during development, right?

rfcs/0001-packaging.md
```diff
@@ -14,33 +14,84 @@ _This RFC does not specify the following planned changes:_
...
+  target: /etc/connector/package-lock.json
+  action: rebuild
+# For any other file change, simply copy it into the existing container and restart it
+- path: ./
+  target: /etc/connector
+  action: sync+restart
+```
+
+## Open Questions
+- Where does the connector declare what type of packaging it uses and its watch configuration?
```
paf31, line 95: On the user's machine, this will be up to the CLI to define, presumably as some sort of configuration plus layout of files on disk.

On Hasura Hub, we can figure this out once we've approved this with the CLI team at a high level, but I expect we'll include a piece of metadata in the hub manifest, plus templated versions of the Dockerfile and any related build input files, in the case of a Dockerfile-based image.

paf31:
@daniel-chambers I'll merge this into my branch, and we can continue work there. Feel free to just push to my branch if you want to change anything, but this looks great.

rfcs/0001-packaging.md
daniel-chambers: Good question, I think we'd definitely want the connectors to provide a default config (I'd almost say it's not optional, but we could just default to the most reliable default: rebuild and restart on any file change), and personally I think it's a good thing if we let users customize that if they wish, but that should absolutely not be required for them to do so.

PR #92, opened 2024-02-02
rfc: rough guidelines for cli plugins

0x777:
[Rendered](https://github.com/hasura/ndc-hub/blob/cli-guidelines/rfcs/0002-cli-guidelines.md)

### DIFF
--- a/rfcs/0002-cli-guidelines.md
+++ b/rfcs/0002-cli-guidelines.md
@@ -71,3 +71,13 @@ could be useful:
    log levels.
    - `HASURA_PLUGIN_CONNECTOR_CONTEXT_PATH` (string) - Fully qualified path to the context
    directory of the connector. 
+
+#### Publishing the Plugin
+   The plugin details (name, version, download location, etc., what we call the plugin manifest) 
+   should be published to the [cli-plugin-index](https://github.com/hasura/cli-plugins-index) repo. 
+   The CLI will track the contents of the `master` branch of this repo and will be able to install
+   any version of the published plugin. (Version should be a CalVer of the format `YYYYMMDD` or a SemVer).
+   Also, we need plugin binaries for `linux/amd64`, `darwin/amd64`, `darwin/arm64`, `windows/amd64` platforms.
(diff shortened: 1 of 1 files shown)


rfcs/0002-cli-guidelines.md
```diff
@@ -0,0 +1,53 @@
...
+could be useful:
+
+1. init (optional)
+
+   Initializing the config directory for a connector. There could be
+   more than one template.
+
+   Let's ignore this for now (out of beta's scope) as the cli's `add
+   ConnectorManifest` would also initialize the connector's config from the
+   connector's default template.
```
daniel-chambers, line 14: This seems like it has the potential to duplicate functionality in the future... which path should the user follow, CLI's `add ConnectorManifest` or this `init`? When should the user choose one over the other?

daniel-chambers: Maybe `add ConnectorManifest` should be built on top of this (ie uses this to provide the init-ing)?

rfcs/0002-cli-guidelines.md
```diff
@@ -0,0 +1,83 @@
...
+   In future, you could also have specific commands to validate parts of the
+   config such as native queries.
+
+4. watch
+
+   This is basically (for the most common use case) the update command run on loop. 
+   This could also be a totally different implementation. This is a convenience 
+   command for the CLI to use during `h3 dev` so that the CLI can invoke this once, 
+   and terminate it once the `dev` command has ended.  
+
```
BenoitRanque, line 56: I'd like to better understand this.
The most naive, straightforward implementation is to literally run `update` in a loop.
Given `update` introspects the database, which may be intensive, this could get quite spammy.

Another option is to loop every X seconds.

A third is to only run the update when a file in the config directory is changed.

None of these seem satisfactory to me. In practice, the user is unlikely to be rapidly modifying their database,
so it's likely not necessary to run the update that often. Changes to the config directory probably don't mean the db has changed, so that option also does not make sense to me.

So, how should this be implemented? Again, happy to literally implement `loop { update() }` as stated here but I really doubt we want that.

PR #94, opened 2024-02-06
RFC: connector deployment spec

paf31:
This is extracted from #89 in order to come to consensus on the aspects of packaging which are relevant only to deployment, while we figure out some of the specifics on the CLI side (watch mode etc.) in the existing RFC.

[Rendered](https://github.com/hasura/ndc-hub/blob/rfc/deployment-spec/rfcs/0000-deployment.md)

---

*Note:* I reverted some of the changes from the other PR:

- I think we agreed that that there is no harm in standardizing the config directory volume mount for all connectors, regardless of how they are built. I prefer this because it's simpler on the SDK side, and connectors can always choose not to use this directory.
- I want to reserve fewer environment variable "namespaces", and env vars can always be remapped in Docker config anyway, so I'm reverting `PORT` to `HASURA_CONNECTOR_PORT`. The SDKs abstract this away from the user.

### DIFF
--- a/rfcs/0000-deployment.md
+++ b/rfcs/0000-deployment.md
@@ -27,13 +27,14 @@ _This RFC does not specify the following planned changes:_
   - `serve` should start a HTTP server on port `8080`, which is compatible with the NDC specification, with `/` as its base URL.
     - For example, `http://connector:8080/query` should implement the query endpoint
     - The default port can be overwritten using the `HASURA_CONNECTOR_PORT` environment variable.
-- The image entrypoint should be set to the connector process, and the default `CMD` should be set to the `serve` command.
+- The image `ENTRYPOINT` should be set to the connector process, and the default `CMD` should be set to the `serve` command. This can mean setting it to the connector executable itself, or some intermediate executable/script that eventually provides its command line arguments to the connector executable.
 - The connector can read environment variables on startup for configuration purposes
   - The following environment variables are reserved, and should not be used for connector-specific configuration:
     - `HASURA_*`
+    - `OTEL_EXPORTER_*`
   - Connectors can use environment variables as part of their configuration. Configuration that varies between different environments or regions (like connection strings) should be configurable via environment variables. 
-- The connector should send any relevant trace spans in the OTLP format to the OTEL collector hosted at the URL provided by the `HASURA_OTLP_ENDPOINT` environment variable.
-  - Spans should indicate the service name provided by the `HASURA_OTEL_SERVICE_NAME` environment variable.	
+- The connector should send any relevant trace spans in the OTLP format to the OTEL collector hosted at the URL provided by the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable.
+  - Spans should indicate the service name provided by the `OTEL_SERVICE_NAME` environment variable.	
 - If the `HASURA_SERVICE_TOKEN_SECRET` environment variable is specified and non-empty, then the connector should implement bearer-token HTTP authorization using the provided static secret token.
 - Information log messages should be logged in plain text to standard output.
 - Error messages should be logged in plain text to standard error.
@@ -41,8 +42,7 @@ _This RFC does not specify the following planned changes:_
 - The connector should respond to the following signals:
   - `SIGTERM`/`SIGINT` - gracefully shutdown the server, and stop the connector process
 - The connector should start as quickly as possible, without any build steps, by reading configuration from disk. Build steps should be performed in the construction of the Docker image, not on startup.
-  - If there is a build step which depends on files controlled by the user (for example, installing dependencies), then they should be moved into a supporting `Dockerfile`. The Docker image which represents the connector would then be the result of `docker build` on this `Dockerfile`. 
-    - To support this use case, tooling should be provided to build a Docker image from a local `Dockerfile`, which may refer to existing Hub connector images.
+  - To support these build steps, tooling should support building images from Dockerfiles. See "Deployment API" below.
 
 ### Open Questions
 
--- a/rfcs/0000-deployment.md
+++ b/rfcs/0000-deployment.md
@@ -1,10 +1,14 @@
-# Connector Packaging
+# Connector Deployment
 
 ## Purpose
 
 For execution of queries and mutations, connectors are specified by the [NDC specification](http://hasura.github.io/ndc-spec/). However, for the purpose of deployment and configuration, their behavior is unspecified, or informally specified. 
 
 This document exists to specify how connectors should be packaged in order to be accepted for inclusion in the Hasura Connector Hub. Any included connectors will be deployable via the CLI.
+
+### Out of Scope
+
+This RFC does not concern itself with the DX aspects of connector metadata authoring, development etc. in the CLI. As it relates to the connector hub, those aspects will be specified in a separate RFC.
 
 ## Related Changes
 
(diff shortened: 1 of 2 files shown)


rfcs/0000-deployment.md
```diff
@@ -0,0 +1,75 @@
...
+## Proposal
+
+- A Hasura Hub data connector will be provided as a Docker image.
+- The connector can expect configuration files to be mounted at `/etc/connector` inside the Docker image on startup. If the `HASURA_CONFIGURATION_DIRECTORY` environment variable is set, it should overwrite this default value.
+  - The connector should not modify these files during execution, and can expect them not to be changed.
+  - The `/etc/connector/.hasura` subdirectory (or `{HASURA_CONFIGURATION_DIRECTORY}/.hasura` in the general case) is reserved for future use and should not be used for configuration. Any connectors which enumerate all subdirectories of `/etc/connector`, for any reason, should ignore this subdirectory if it exists.
+- The connector executable should accept the following subcommands:
+  - `serve` should start a HTTP server on port `8080`, which is compatible with the NDC specification, with `/` as its base URL.
+    - For example, `http://connector:8080/query` should implement the query endpoint
+    - The default port can be overwritten using the `HASURA_CONNECTOR_PORT` environment variable.
```
daniel-chambers, line 29: It was suggested by @SamirTalwar on the other RFC (and was changed there) that we use the standard `PORT` env var.

rfcs/0000-deployment.md
```diff
@@ -0,0 +1,75 @@
...
+  },
+  FromDockerfileAndBuildInputs {
+    tar: TarFile,
+  }
+}
+```
+
+How this structure gets built by CLI (or its supporting web service) is out of scope. For example, we might fetch tar bundles from Git repos, or from the filesystem. Dockerfiles might be under the user's control, or not. But this structure is what is required to build images for deployment.
+
+In the case of `FromHubConnector`, the expectation is that the connector build service maintains a list of prebuilt images, indexed by the names and versions of hub connectors.
```
daniel-chambers, line 73: Does this mean "Hub Connectors" must only use pre-built images? If so, it might be wise to specialize that FromHubConnector name to FromPrebuiltHubConnector, given that we may have "Hub Connectors" that require building, and these would be addressed by using `FromDockerfileAndBuildInputs` (I guess?).

rfcs/0000-deployment.md
daniel-chambers: Oh, sorry I missed your comment in the PR description on this. This isn't a hill I'm willing to die on, but I will point out that PORT is often automatically supported by HTTP servers.

rfcs/0000-deployment.md
```diff
@@ -0,0 +1,75 @@
...
+- There will also be no more `HasuraHubConnector` in v3-engine metadata. Instead the engine will only see connector URLs, and the CLI will manage the instantiation and deployment of connectors, and the creation of those URLs.
+
+## Proposal
+
+- A Hasura Hub data connector will be provided as a Docker image.
+- The connector can expect configuration files to be mounted at `/etc/connector` inside the Docker image on startup. If the `HASURA_CONFIGURATION_DIRECTORY` environment variable is set, it should overwrite this default value.
+  - The connector should not modify these files during execution, and can expect them not to be changed.
+  - The `/etc/connector/.hasura` subdirectory (or `{HASURA_CONFIGURATION_DIRECTORY}/.hasura` in the general case) is reserved for future use and should not be used for configuration. Any connectors which enumerate all subdirectories of `/etc/connector`, for any reason, should ignore this subdirectory if it exists.
+- The connector executable should accept the following subcommands:
+  - `serve` should start a HTTP server on port `8080`, which is compatible with the NDC specification, with `/` as its base URL.
```
daniel-chambers, line 27: @SamirTalwar added a comment to the old PR about this: https://github.com/hasura/ndc-hub/pull/89#discussion_r1480143588

rfcs/0000-deployment.md
paf31: No, this name just needs to be changed. Perhaps it can just be called `FromDockerImage`.

rfcs/0000-deployment.md
paf31: Environment variables are exposed by the connector package though, they are part of the specification of a connector package that the user will see. 

Environment variables can be remapped in Dockerfiles if need be, but if we pick `PORT` then it's another name we need to reserve. A connector author may want to use `PORT` to mean "the port to contact the database on", but they're unlikely to ever want to call something `HASURA_CONNECTOR_{anything}`.

rfcs/0000-deployment.md
paf31: If we ever add optional standard arguments to `--serve` then it becomes more useful to be able to assert it exists now.

SamirTalwar: I'm not generally comfortable with arguments of the form, "we may want to do some other thing in the future, so this is a good idea".

Firstly, this is the first I'm hearing of these standard arguments, and I am not convinced it's a good idea. (Environment variables tend to be more composable.)

Secondly, we can always introduce the concept later if we need it, in a staggered release fashion.

rfcs/0000-deployment.md
SamirTalwar: I'm alright with this argument.

rfcs/0000-deployment.md
```diff
@@ -0,0 +1,75 @@
...
+    - For example, `http://connector:8080/query` should implement the query endpoint
+    - The default port can be overwritten using the `HASURA_CONNECTOR_PORT` environment variable.
+- The image `ENTRYPOINT` should be set to the connector process, and the default `CMD` should be set to the `serve` command. This can mean setting it to the connector executable itself, or some intermediate executable/script that eventually provides its command line arguments to the connector executable.
+- The connector can read environment variables on startup for configuration purposes
+  - The following environment variables are reserved, and should not be used for connector-specific configuration:
+    - `HASURA_*`
+    - `OTEL_EXPORTER_*`
+  - Connectors can use environment variables as part of their configuration. Configuration that varies between different environments or regions (like connection strings) should be configurable via environment variables.
```
SamirTalwar, line 35: Is there a "namespace" for these? Perhaps `CONNECTOR_*`?

rfcs/0000-deployment.md
```diff
@@ -0,0 +1,75 @@
...
+- The connector should send any relevant trace spans in the OTLP format to the OTEL collector hosted at the URL provided by the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable.
+  - Spans should indicate the service name provided by the `OTEL_SERVICE_NAME` environment variable.	
+- If the `HASURA_SERVICE_TOKEN_SECRET` environment variable is specified and non-empty, then the connector should implement bearer-token HTTP authorization using the provided static secret token.
+- Information log messages should be logged in plain text to standard output.
+- Error messages should be logged in plain text to standard error.
+- On startup, in case of failure to start, the connector should flush any error messages to standard error, and exit with a non-zero exit code.
+- The connector should respond to the following signals:
+  - `SIGTERM`/`SIGINT` - gracefully shutdown the server, and stop the connector process
+- The connector should start as quickly as possible, without any build steps, by reading configuration from disk. Build steps should be performed in the construction of the Docker image, not on startup.
```
0x777, line 44: Do you think we should include the motivation here? 

> A connector process could be spun up to serve a request.

0x777:
LGTM

rfcs/0000-deployment.md
paf31: Thinking about this some more, I think you're right, and we can simplify things a bit the way you suggested. I'll change it, thanks.

rfcs/0000-deployment.md
paf31: Not currently, but maybe there should be. I was actually thinking that we could remove any reserved variables by using command line args, and leave all environment variables available for use. But OpenTelemetry stops us from doing that, since they require env vars.

daniel-chambers (approved):
LGTM!

rfcs/0000-deployment.md
SamirTalwar: Perhaps we can leave this for now and add it as a "best practice" if we feel the need.

SamirTalwar (approved):

rfcs/0000-deployment.md
```diff
@@ -0,0 +1,75 @@
...
+  - For example, `http://connector:8080/query` should implement the query endpoint
+  - The default port can be overwritten using the `HASURA_CONNECTOR_PORT` environment variable.
+  - This can mean setting `ENTRYPOINT` to the connector executable itself, or some intermediate executable/script that eventually provides its command line arguments to the connector executable.
+- The connector can read environment variables on startup for configuration purposes
+  - The following environment variables are reserved, and should not be used for connector-specific configuration:
+    - `HASURA_*`
+    - `OTEL_EXPORTER_*`
+  - Connectors can use environment variables as part of their configuration. Configuration that varies between different environments or regions (like connection strings) should be configurable via environment variables. 
+- The connector should send any relevant trace spans in the OTLP format to the OTEL collector hosted at the URL provided by the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable.
```
BenoitRanque, line 35: This should be changed, per the [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp) the `OTEL_EXPORTER_OTLP_ENDPOINT` is a base url, and
> For the per-signal variables (OTEL_EXPORTER_OTLP_<signal>_ENDPOINT), the URL MUST be used as-is without any modification. The only exception is that if an URL contains no path part, the root path / MUST be used (see [Example > 2](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#example-2)).

 
> If signals are sent that have no per-signal configuration from the previous point, OTEL_EXPORTER_OTLP_ENDPOINT is used as a base URL and the signals are sent to these paths relative to that:

 
> Traces: v1/traces
> Metrics: v1/metrics.
> Logs: v1/logs.

So the URL for traces would be specified with either `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` or `<OTEL_EXPORTER_OTLP_ENDPOINT>/v1/traces`

paf31: > to the OTEL collector hosted at the URL provided by the `OTEL_EXPORTER_OTLP_ENDPOINT`

I think this is compatible with what the SDK does, and what the spec says? The OTEL collector _root_ URL is provided in the `OTEL_EXPORTER_OTLP_ENDPOINT` variable. I will add a sentence clarifying this.

PR #95, opened 2024-02-12
CI: Copy Cargo.lock into the build image to pin dependencies.

SamirTalwar:
Otherwise we're not testing what we expect.


sordina (approved):
Good change.

PR #96, opened 2024-02-13
Remove constraints from `default_main` where possible.

SamirTalwar:
I have made a few changes here.

1. I moved `Sync + Send` to the connector type constraints, because they're so basic that I can't imagine a connector working without them.
2. I removed the constraint on connectors themselves being cloneable by manually implementing `Clone` on `ServerState`.
3. I removed all unused constraints that I could find.

### DIFF
--- a/rust-connector-sdk/src/default_main.rs
+++ b/rust-connector-sdk/src/default_main.rs
@@ -128,11 +128,25 @@ struct CheckHealthCommand {
 
 type Port = u16;
 
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct ServerState<C: Connector> {
     configuration: C::Configuration,
     state: C::State,
     metrics: Registry,
+}
+
+impl<C: Connector> Clone for ServerState<C>
+where
+    C::Configuration: Clone,
+    C::State: Clone,
+{
+    fn clone(&self) -> Self {
+        Self {
+            configuration: self.configuration.clone(),
+            state: self.state.clone(),
+            metrics: self.metrics.clone(),
+        }
+    }
 }
 
 /// A default main function for a connector.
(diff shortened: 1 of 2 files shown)


rust-connector-sdk/src/default_main.rs
```diff
@@ -158,12 +172,12 @@ pub struct ServerState<C: Connector> {
...
 /// - It reports traces to an OTLP collector specified on the command line,
 /// - Logs are written to stdout
-pub async fn default_main<C: Connector + Clone + Default + 'static>(
+pub async fn default_main<C: Connector + Default + 'static>(
 ) -> Result<(), Box<dyn Error + Send + Sync>>
 where
-    C::RawConfiguration: Serialize + DeserializeOwned + JsonSchema + Sync + Send,
-    C::Configuration: Serialize + DeserializeOwned + Sync + Send + Clone,
-    C::State: Sync + Send + Clone,
+    C::RawConfiguration: Serialize + DeserializeOwned + JsonSchema,
```
plcplc, line 178: Taking the argument of this PR a bit further, is there any sense to any `impl Connector for C` that doesn't satisfy the constraints of `default_main`?

A connector that doesn't might as well not use ndc-sdk at all.

SamirTalwar: Well, you can make your own `main`, in theory. I was wondering this but didn't want to push it too far.

rust-connector-sdk/src/default_main.rs
```diff
@@ -176,13 +190,13 @@ where
     }
 }
 
-async fn serve<C: Connector + Clone + Default + 'static>(
+async fn serve<C: Connector + Default + 'static>(
```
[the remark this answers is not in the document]
SamirTalwar: Good catch! Pushed.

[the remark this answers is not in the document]
SamirTalwar: We can also remove `'static' once we remove the configuration server.

rust-connector-sdk/src/default_main.rs
plcplc: Yeah, maybe. I'm just wondering what if anything anyone who makes their own `main` function actually gets from using the sdk.

At any rate this is still a fine improvement.

plcplc (approved):

PR #97, opened 2024-02-13
Update SDK for new deployment spec

paf31:
- Use new environment variable names and defaults
- Remove configuration server
- Pass `PathBuf` to `validate_raw_configuration` and remove `RawConfiguration`
- Update `Dockerfile` to meet the deployment spec

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,7 @@ RUN apt-get update \
 ENV CARGO_HOME=/app/.cargo
 ENV RUSTFLAGS="-C link-arg=-fuse-ld=lld"
 
+COPY Cargo.lock .
 COPY ./rust-connector-sdk .
 
 RUN cargo build --release
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,4 +23,4 @@ RUN set -ex; \
       libssl-dev
 COPY --from=build /app/target/release/ndc_hub_example ./ndc_hub_example
 ENTRYPOINT [ "/ndc_hub_example" ]
-CMD [ "serve" ]
+CMD [ "serve", "--configuration", "/etc/connector" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,6 +16,10 @@ COPY ./rust-connector-sdk .
 RUN cargo build --release
 
 FROM debian:buster-slim as connector
+RUN apt-get update \
+ && DEBIAN_FRONTEND=noninteractive \
+    apt-get install --no-install-recommends --assume-yes\
+    libssl-dev
 COPY --from=build /app/target/release/ndc_hub_example ./ndc_hub_example
 ENTRYPOINT [ "/ndc_hub_example" ]
-CMD [ "serve", "--port", "8080" ]
+CMD [ "serve" ]
--- a/rust-connector-sdk/src/connector.rs
+++ b/rust-connector-sdk/src/connector.rs
@@ -1,4 +1,4 @@
-use std::error::Error;
+use std::{error::Error, path::PathBuf};
 
 use async_trait::async_trait;
 use ndc_client::models;
@@ -30,16 +30,6 @@ pub struct InvalidRange {
 pub enum KeyOrIndex {
     Key(String),
     Index(u32),
-}
-
-/// Errors which occur when trying to validate connector
-/// configuration.
-///
-/// See [`Connector::update_configuration`].
-#[derive(Debug, Error)]
-pub enum UpdateConfigurationError {
-    #[error("error validating configuration: {0}")]
-    Other(#[from] Box<dyn Error + Send + Sync>),
 }
 
 /// Errors which occur when trying to initialize connector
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -16,6 +16,10 @@ COPY ./rust-connector-sdk .
 RUN cargo build --release
 
 FROM debian:buster-slim as connector
+RUN apt-get update \
+ && DEBIAN_FRONTEND=noninteractive \
+    apt-get install --no-install-recommends --assume-yes\
+    libssl-dev
```
SamirTalwar, line 22: Adding `set -ex` makes it easier to figure out which line broke when something goes wrong.

```suggestion
RUN set -ex; \
    apt-get update; \
    DEBIAN_FRONTEND=noninteractive \
      apt-get install --no-install-recommends --assume-yes \
      libssl-dev
```

rust-connector-sdk/src/connector.rs
```diff
@@ -201,23 +191,15 @@ pub enum MutationError {
...
-
-    async fn update_configuration(
-        config: Self::RawConfiguration,
-    ) -> Result<Self::RawConfiguration, UpdateConfigurationError>;
-
     /// Validate the raw configuration provided by the user,
     /// returning a configuration error or a validated [`Connector::Configuration`].
     async fn validate_raw_configuration(
-        configuration: Self::RawConfiguration,
+        configuration_dir: PathBuf,
```
SamirTalwar, line 202: I suggest weakening this to `AsRef<Path>` so we can be a little more flexible with the input.

```suggestion
        configuration_dir: impl AsRef<Path>,
```

rust-connector-sdk/src/default_main.rs
```diff
@@ -243,17 +233,13 @@ where
 
 /// Initialize the server state from the configuration file.
 pub async fn init_server_state<C: Connector + Clone + Default + 'static>(
-    config_file: impl AsRef<Path>,
+    config_directory: PathBuf,
```
SamirTalwar, line 236:
```suggestion
    config_directory: impl AsRef<Path>,
```

SamirTalwar (approved):
Approving so it's not blocked, but please consider my comments.

rust-connector-sdk/src/default_main.rs
paf31: It would have to be `AsRef<Path> + Send` apparently.

SamirTalwar: Fair; that's OK with me if it's OK with you. Looks like that's fine for `PathBuf` and `String`.

SamirTalwar (approved):

rust-connector-sdk/src/connector.rs
```diff
@@ -201,23 +191,15 @@ pub enum MutationError {
...
 
-    fn make_empty_configuration() -> Self::RawConfiguration;
-
-    async fn update_configuration(
-        config: Self::RawConfiguration,
-    ) -> Result<Self::RawConfiguration, UpdateConfigurationError>;
-
     /// Validate the raw configuration provided by the user,
     /// returning a configuration error or a validated [`Connector::Configuration`].
     async fn validate_raw_configuration(
```
SamirTalwar, line 201: While we're here, how about renaming this to `parse_configuration`?

rust-connector-sdk/src/default_main.rs
```diff
@@ -60,69 +53,68 @@ enum Command {
...
 #[derive(Clone, Parser)]
 struct TestCommand {
     #[arg(long, value_name = "SEED", env = "SEED")]
     seed: Option<String>,
-    #[arg(long, value_name = "CONFIGURATION_FILE", env = "CONFIGURATION_FILE")]
+    #[arg(
+        long,
+        value_name = "DIRECTORY",
+        env = "HASURA_CONFIGURATION_DIRECTORY",
+        default_value = "/etc/connector"
```
daniel-chambers, line 88: I'm not sure how easy this is to do, but is it possible to remove this default when compiled for Windows? I know we don't care about deployment on Windows, but theoretically Windows developers may build a connector using this SDK and when they run it locally, it should not default to an impossible path, instead we should prefer to just make them specify this manually.

daniel-chambers: This also applies to the other commands that have this same parameter: ServeCommand, ReplayCommand

SamirTalwar: Hmmm, I think this point is bigger than just Windows vs. Unices.

In a Linux Docker image, we want to read `/etc/connector`. But when running locally for testing, we want to read the specified directory; we almost never want to use `/etc/connector`, even on Linux.

I think the right move here is to remove the default value and set `ENV HASURA_CONFIGURATION_DIRECTORY /etc/connector` in connector Docker images. This can be overridden.

Long story short, the default is a packaging concern, not an SDK concern.

paf31: As I think about it, I agree. I'll remove the default and move it to the Dockerfile. For reference, it was possible though:

```rust
#[cfg_attr(target_family = "unix", arg(default_value = "/etc/connector"))]
```

SamirTalwar (approved):

SamirTalwar:
Let's merge this. Gonna follow up with a change to an error enum but that needs its own review.

PR #98, opened 2024-02-15
Distribution RFC

sordina:


rfcs/0002-distribution.md
```diff
@@ -0,0 +1,249 @@
...
+
+* [NDC specification](http://hasura.github.io/ndc-spec/)
+* [Deployment Specification](https://github.com/hasura/ndc-hub/blob/main/rfcs/0000-deployment.md)
+* [Packaging Specification (WIP)](https://github.com/hasura/ndc-hub/pull/89/files)
+
+This new distribution specification details how connector packages are intended to be owned, stored, indexed, searched, fetched and automatically published.
+
+The intuition for this system is inspired by other package management systems such as NPM, Cabal, etc.
+
+There was a previous implementation of these concepts as described (TODO: Get docs links from Shraddha)
```
scriptnull, line 18: we might be working on the v2 of this legacy implementation. We could probably try to accommodate this spec in this v2 implementation.

cc: @imperfect-fourth

sordina: @scriptnull 👍

rfcs/0002-distribution.md
```diff
@@ -0,0 +1,249 @@
...
+* Request for package download URI
+
+
+### Applications / CLI
+
+The Hasura V3 CLI will provide a consistent and convenient interface to interacting with the API.
+
+The CLI commands will include the following:
+
+* hasura3 package create
```
scriptnull, line 154: I think in the new DX implementation, we always have `hasura3 <verb> <noun>`. So, please consult with DX team to see if this is okay.

rfcs/0002-distribution.md
```diff
@@ -0,0 +1,247 @@
...
+* hasura3 package publish
+* hasura3 package revoke
+* hasura3 package search
+* hasura3 package fetch
```
scriptnull, line 158: Should we aim to include the word `connector` somehow in the package-related commands?  Because package might be something that we want the user to relate to Hasura connectors rather than it being like a top-level concept in Hasura.

sordina: Good idea! Package could be implicit by the operations on the connector.

paf31:
Can you please add some notes about the motivation here, and specifically why we're considering building something new, as opposed to the Git-based approach that worked for the v2 GDC registry? My concern is primarily that a PR-based workflow is a lot simpler for connector authors than learning and using some custom API or CLI.

sordina:
@paf31 I've added a `Motivation and Why and How this Differs from Existing Solutions` section.

The main reason to change the existing system is simply that if we're packaging definitions independently of the git registry we should have somewhere to index/reference them. The rest follows from there. That being said if people find the registry workflows convenient there's no reason why this couldn't be preserved as a tributary to the system.

The current system already has the DB/API components mentioned so it's not that radically different in practice.

What do you think?

rfcs/0002-distribution.md
```diff
@@ -48,6 +48,32 @@ The following are not described in this specification:
...
+
+The current system's GraphQL: https://data.pro.arusah.com/ - Looking at all the graphql queries starting with connector_ should give all the things that we have so far.
+
+Google cloud function (https://github.com/hasura/connectors-cloud-integration/tree/main/sync_connector_hub_data) runs every 24 hours and scrapes registry and inserts into DB.
+
+Issues with the current system:
+
+* Pull-based ingestion of registry - Changes should ideally propagate instantaneously
+* No ability to treat connectors independently - Need to PR to central registry
+* Artefacts are not distributable outside of Github references - Tied to Github
```
paf31, line 73: I think there's a simple workaround for this - link to the release tarball on GitHub, or any other tarball hosted anywhere else.

rfcs/0002-distribution.md
```diff
@@ -48,6 +48,32 @@ The following are not described in this specification:
...
+| Third-party CI | No current component | Usage via CLI/API | Third-parties are able to integrate into their CI via API |
+| Topic Scraping | No current component | Scheduled/Webhook trigger of topic ingestion of community connectors | Arbitrary ingestion is possible via the API/CLI |
+
+The current system's GraphQL: https://data.pro.arusah.com/ - Looking at all the graphql queries starting with connector_ should give all the things that we have so far.
+
+Google cloud function (https://github.com/hasura/connectors-cloud-integration/tree/main/sync_connector_hub_data) runs every 24 hours and scrapes registry and inserts into DB.
+
+Issues with the current system:
+
+* Pull-based ingestion of registry - Changes should ideally propagate instantaneously
```
paf31, line 71: Can't we fix this by running the job more frequently, or in a commit hook?

paf31:
@sordina Thanks. I've added a couple of notes on two of the three points you noted. On the third:

> No ability to treat connectors independently - Need to PR to central registry

I think this is a feature, not a bug. If the registry is considered as a community-driven resource, then I think a PR is the appropriate way to suggest and review changes. Otherwise, we have to build an awful lot of custom versions of open tools.

sordina:
@paf31   How about if the existing behaviour is preserved in the new system as one tributary? The new proposed system very closely matches the existing system anyway.

PR #102, opened 2024-02-16
Rename and extend `ValidateError`.

SamirTalwar:
We now need to conduct I/O during `parse_configuration`, which means the current `ValidateError` isn't quite enough.

I've renamed it to `ParseError` (to complement `parse_configuration`), and added a few extra cases.

1. `ParseError`, which denotes the file path and position where parsing failed.
2. `ValidateError`, as before, but including the file path, wrapping the vector in `InvalidNodes` so we can implement `Display`.
3. `IoError`, for when we fail to read a file.
4. `Other`, because I can't think of everything.

I am using the first 3 in integrating the latest changes into ndc-postgres.

### DIFF
--- a/rust-connector-sdk/src/connector.rs
+++ b/rust-connector-sdk/src/connector.rs
@@ -18,7 +18,7 @@ pub mod example;
 #[derive(Debug, Error)]
 pub enum ParseError {
     #[error("error parsing configuration: {0}")]
-    ParseError(Position),
+    ParseError(LocatedError),
     #[error("error validating configuration: {0}")]
     ValidateError(InvalidPaths),
     #[error("error processing configuration: {0}")]
@@ -29,20 +29,22 @@ pub enum ParseError {
 
 /// The position of a single character in a text file.
 #[derive(Debug, Clone)]
-pub struct Position {
+pub struct LocatedError {
     pub path: PathBuf,
     pub line: usize,
     pub column: usize,
+    pub message: String,
 }
 
-impl Display for Position {
+impl Display for LocatedError {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(
             f,
-            "{0}:{1}:{2}",
+            "{0}:{1}:{2}: {3}",
             self.path.display(),
             self.line,
-            self.column
+            self.column,
+            self.message
         )
     }
 }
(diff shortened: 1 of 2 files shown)


rust-connector-sdk/src/connector.rs
```diff
@@ -12,17 +14,69 @@ pub mod example;
...
 ///
-/// See [`Connector::validate_raw_configuration`].
+/// See [`Connector::parse_configuration`].
 #[derive(Debug, Error)]
-pub enum ValidateError {
-    #[error("error validating configuration: {0:?}")]
-    ValidateError(Vec<InvalidRange>),
+pub enum ParseError {
+    #[error("error parsing configuration: {0}")]
+    ParseError(Position),
```
soupi, line 21: Don't we also need the actual error message here?

SamirTalwar: FYI, I'm using this as follows (now I've added the message):

```rust
serde_json::from_reader(configuration_reader).map_err(|error| {
  connector::ParseError::ParseError(connector::LocatedError {
    path: configuration_file,
    line: error.line(),
    column: error.column(),
    message: error.to_string(),
  })
})

File: rust-connector-sdk/src/connector.rs [hunk 2]
```diff
@@ -12,17 +14,71 @@ pub mod example;
 /// Errors which occur when trying to validate connector
 /// configuration.
 ///
-/// See [`Connector::validate_raw_configuration`].
+/// See [`Connector::parse_configuration`].
 #[derive(Debug, Error)]
-pub enum ValidateError {
-    #[error("error validating configuration: {0:?}")]
-    ValidateError(Vec<InvalidRange>),
```
paf31, line 19: There were previously many invalid ranges, but now only one?

SamirTalwar: It's now an `InvalidPaths` object, which wraps `Vec<InvalidPath>` (renamed from `InvalidRange`).

I wrapped it so I could add `impl Display`.

soupi (approved):

rust-connector-sdk/src/connector.rs
paf31: We can still have multiple errors in many files, and we'd want to presumably highlight them all at once in an editor errors pane, for example (if we ended up standardizing a plugin subcommand for this purpose)

SamirTalwar: Seems legit; perhaps we should add a `path` here too.

soupi (approved):
lgtm.

PR #106, opened 2024-02-21
RFC | add more environment variables to OTEL exporter

hgiasac:
- Add more required configurations for OTEL exporter (`OTEL_EXPORTER_OTLP_HEADERS` and `OTEL_RESOURCE_ATTRIBUTES`)
- Relax other OTEL configs to follow the [OTEL spec](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/). Existing OTEL exporter SDKs mostly support those environment variables internally. Therefore we don't need to implement them in connector SDKs.

[Rendered](https://github.com/hasura/ndc-hub/blob/toan/add-otel-headers-env/rfcs/0000-deployment.md#proposal)

### DIFF
--- a/rfcs/0000-deployment.md
+++ b/rfcs/0000-deployment.md
@@ -35,6 +35,9 @@ _This RFC does not specify the following planned changes:_
 - The connector should send any relevant trace spans in the OTLP format to the OTEL collector hosted at the URL provided by the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable. 
   - Note: `OTEL_EXPORTER_OTLP_ENDPOINT` indicates the _root URL_ of the collector, and not the `/v1/traces` endpoint.
   - Spans should indicate the service name provided by the `OTEL_SERVICE_NAME` environment variable.	
+  - The connector can configure additional headers to add in outgoing gRPC or HTTP requests by the `OTEL_EXPORTER_OTLP_HEADERS` environment variables.
+  - The connector also can configure additional resource attributes for requests by the `OTEL_RESOURCE_ATTRIBUTES` environment variables.
+  - Note: the value of `OTEL_EXPORTER_OTLP_HEADERS` and `OTEL_RESOURCE_ATTRIBUTES` is key-value pairs, for example, `api-key=key,other-config-value=value`.
 - If the `HASURA_SERVICE_TOKEN_SECRET` environment variable is specified and non-empty, then the connector should implement bearer-token HTTP authorization using the provided static secret token.
 - Information log messages should be logged in plain text to standard output.
 - Error messages should be logged in plain text to standard error.
--- a/rfcs/0000-deployment.md
+++ b/rfcs/0000-deployment.md
@@ -31,7 +31,6 @@ _This RFC does not specify the following planned changes:_
   - The following environment variables are reserved, and should not be used for connector-specific configuration:
     - `HASURA_*`
     - `OTEL_*`
-    - `OTEL_EXPORTER_*`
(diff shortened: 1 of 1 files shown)


rfcs/0000-deployment.md
```diff
@@ -30,11 +30,15 @@ _This RFC does not specify the following planned changes:_
...
 - The connector should send any relevant trace spans in the OTLP format to the OTEL collector hosted at the URL provided by the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable. 
   - Note: `OTEL_EXPORTER_OTLP_ENDPOINT` indicates the _root URL_ of the collector, and not the `/v1/traces` endpoint.
   - Spans should indicate the service name provided by the `OTEL_SERVICE_NAME` environment variable.	
+  - The connector can configure additional headers to add in outgoing gRPC or HTTP requests by the `OTEL_EXPORTER_OTLP_HEADERS` environment variables.
+  - The connector also can configure additional resource attributes for requests by the `OTEL_RESOURCE_ATTRIBUTES` environment variables.
+  - Note: the value of `OTEL_EXPORTER_OTLP_HEADERS` and `OTEL_RESOURCE_ATTRIBUTES` is key-value pairs, for example, `api-key=key,other-config-value=value`.
+  - Other OTLP exporter configurations are optional. However, they must follow the [Environment Variable Specification](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables) and [SDK config](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/).
```
paf31, line 41: I think we need to be a bit careful here because the environment variable spec says too much IMO. For example, it requires `OTEL_PROPAGATORS`, but the NDC spec requires `tracecontext` propagation. There are a lot of other env vars that we don't need to support in full.

paf31 (commented):
I'm wondering if we should specify a list of individual env vars to support, or to just say "a connector should support OTEL in full" but clarify which parts we actually care about right now.

rfcs/0000-deployment.md
```diff
@@ -30,11 +30,15 @@ _This RFC does not specify the following planned changes:_
 - The connector can read environment variables on startup for configuration purposes
   - The following environment variables are reserved, and should not be used for connector-specific configuration:
     - `HASURA_*`
-    - `OTEL_EXPORTER_*`
+    - `OTEL_*`
   - Connectors can use environment variables as part of their configuration. Configuration that varies between different environments or regions (like connection strings) should be configurable via environment variables. 
 - The connector should send any relevant trace spans in the OTLP format to the OTEL collector hosted at the URL provided by the `OTEL_EXPORTER_OTLP_ENDPOINT` environment variable. 
   - Note: `OTEL_EXPORTER_OTLP_ENDPOINT` indicates the _root URL_ of the collector, and not the `/v1/traces` endpoint.
   - Spans should indicate the service name provided by the `OTEL_SERVICE_NAME` environment variable.	
+  - The connector can configure additional headers to add in outgoing gRPC or HTTP requests by the `OTEL_EXPORTER_OTLP_HEADERS` environment variables.
```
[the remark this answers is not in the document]
hgiasac: You're right. My words may cause confusing here

rfcs/0000-deployment.md
hgiasac: I have added a line to clarify the required `tracecontext` propagation.

> I'm wondering if we should specify a list of individual env vars to support, or to just say "a connector should support OTEL in full" but clarify which parts we actually care about right now.

IMO not [all languages](https://opentelemetry.io/docs/languages/) implement OTEL SDK. It seems too much to expect the author to support OTEL in full. However, we can clarify that the connector can support other configurations depending on the underlying SDK.

Anyway, it would be great if you could suggest better words to make the proposal clearer.

PR #108, opened 2024-02-23
Add an `InitializationContext`.

SamirTalwar:
The initializion context is a value of a type specified by the connector, which can be injected into the entrypoint. This allows users of the connector functions to provide a little more control over how the connector configures itself.

### DIFF
--- a/rust-connector-sdk/src/connector.rs
+++ b/rust-connector-sdk/src/connector.rs
@@ -221,33 +221,27 @@ pub enum MutationError {
 
 /// Connectors using this library should implement this trait.
 ///
+/// It provides methods which implement the standard endpoints defined by the specification:
+/// capabilities, schema, query, mutation, query/explain, and mutation/explain.
 ///
-/// It provides methods which implement the standard endpoints
-/// defined by the specification: capabilities, schema, query, mutation,
-/// query/explain, and mutation/explain.
-///
-/// In addition, it introduces names for types to manage
-/// state and configuration (if any), and provides any necessary context
-/// for observability purposes (metrics, logging and tracing).
+/// In addition, it introduces names for types to manage state and configuration (if any), and
+/// provides any necessary context for observability purposes (metrics, logging and tracing).
 ///
 /// ## Configuration
 ///
-/// Connectors encapsulate data sources, and likely require configuration
-/// (connection strings, web service tokens, etc.). The NDC specification
-/// does not discuss this sort of configuration, because it is an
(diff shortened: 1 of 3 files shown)


rust-connector-sdk/src/connector.rs
```diff
@@ -221,54 +221,50 @@ pub enum MutationError {
...
-/// would be configuration, but a connection pool object created from that
-/// connection string would be state.
+/// State is distinguished from configuration in that it is not provided directly by the user, and
+/// is transient. For example, a connection string would be configuration, but a connection pool
+/// object created from that connection string would be state.
 #[async_trait]
 pub trait Connector {
-    /// The type of validated configuration
+    /// Context used to initialize the server state
+    type InitializationContext: Sync + Send;
```
paf31, line 258: How are you supposed to use `Connector` abstractly if `InitializationContext` only appears in input position? Why not pass this data into the constructor of the implementing data type?

SamirTalwar: I tried that first but it wasn't great for two reasons:

1. I have to add a reference to the connector object everywhere, whereas now it's entirely a phantom type.
2. That information becomes available during request, and I only want it to be available at initialization time.

We could potentially make `InitializationContext` an associated type implementing a new trait, `ConnectorInitialization`, and move the `parse_configuration` and `try_init_state` methods into that trait.

As for how we use it abstractly, it's an input to `default_main_with`, which means the SDK just needs to know it exists, and not what it is.

dmoverton (approved):

rust-connector-sdk/src/connector.rs
SamirTalwar: What do you think of the alternative in #111?

dmoverton: I hadn't seen #111 . It is probably a bit cleaner to separate it out into a new trait, but I think I prefer that.

daniel-chambers: Personally, I prefer the approach in this PR. It is simpler, and I don't see a lot of value add that #111 brings. But they're so similar that I don't really care which approach is used. Just pick one and smash it in I reckon 🙂

SamirTalwar: I think I prefer #111, so if you feel like approving that, please do. 😁

SamirTalwar: Cool, closing this one, thanks!

PR #113, opened 2024-02-28
Add a  specific error variant for missing configuration files

i-am-tom:
I experienced a very unhelpful error when I pointed to a configuration directory that _didn't_ contain a `configuration.json`. I'd like to add an error variant for this, and then use this variant in `ndc-postgres` when we update its version of `ndc-sdk`.

### DIFF
--- a/rust-connector-sdk/src/connector.rs
+++ b/rust-connector-sdk/src/connector.rs
@@ -21,6 +21,8 @@ pub enum ParseError {
     ParseError(LocatedError),
     #[error("error validating configuration: {0}")]
     ValidateError(InvalidNodes),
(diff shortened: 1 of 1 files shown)


SamirTalwar (approved):

SamirTalwar:
I just thought about this some more, and I think it might be worth updating the `IoError` case to also wrap an `Option<PathBuf>` (and deleting the `From` instance), or creating a new `FileError { path: PathBuf, #[source] source: io::Error }`.

There are lots of things that can go wrong when reading files, and if the path is known, I think it's worth capturing it in all instances, not just this one.

I am totally fine with this not happening right now.

PR #116, opened 2024-03-07
Delete the v2 compatibility routes.

SamirTalwar:
It seems that no one is using these any more.


hallettj (approved):

BenoitRanque (approved):
Glad to see this go

BenoitRanque:
I think we can also remove the [dependency on the v2 rust types](https://github.com/hasura/ndc-hub/blob/main/rust-connector-sdk/Cargo.toml#L19)

paf31:
This is referenced here: https://hasura.io/docs/3.0/connectors/v2-compatibility

How are we making these decisions to remove features?

SamirTalwar:
I asked around if anyone was using this and did not get any affirmatives, or anyone pointing out that it's useful. I didn't know about the docs.

Happy to restore them if they're useful. I will put them under a Cargo feature so we can avoid the `gdc_rust_types` dependency in ndc-postgres, and update the docs accordingly.

PR #125, opened 2024-03-18
Add NodeJS Lambda connector metadata

daniel-chambers:
Adds the metadata for the NodeJS Lambda connector to the hub registry.


sordina (approved):
Do we need to include the previous package releases?

daniel-chambers:
@sordina I don't see why not. I expect they're already in the DB, so having them here will make it easier to sync that with this once we get to that.

sordina:
Ok sounds good.

PR #129, opened 2024-03-22
Dockerized Commands in the packaging spec

daniel-chambers:
[Rendered](https://github.com/hasura/ndc-hub/blob/daniel/dockerized-commands-rfc/rfcs/0001-packaging.md)

### DIFF
--- a/rfcs/0001-packaging.md
+++ b/rfcs/0001-packaging.md
@@ -64,8 +64,14 @@ type EnvironmentVariableDefinition = {
 }
 
 type Commands = {
-  update?: string
-  watch?: string
+  update?: string | DockerizedCommand
+  watch?: string | DockerizedCommand
+}
+
+type DockerizedCommand = {
+  type: "Dockerized"
+  dockerImage: string // eg "hasura/postgres-data-connector:1.0.0"
+  commandArgs: string[]
 }
 
 type CliPluginDefinition = {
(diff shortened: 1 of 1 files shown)


rfcs/0001-packaging.md
```diff
@@ -64,8 +64,14 @@ type EnvironmentVariableDefinition = {
 }
 
 type Commands = {
-  update?: string
-  watch?: string
+  update?: string | DockerizedCommand
+  watch?: string | DockerizedCommand
```
SamirTalwar, line 68: If we're going to allow a string array for args in the Docker command, let's do it for native commands too.

```suggestion
  update?: string | string[] | DockerizedCommand
  watch?: string | string[] | DockerizedCommand
```

daniel-chambers: Hmm, I'm not sure about that. For native commands, that is not "args", it's a shell command, so it's one string because it ends up as one arg. It's what get passed to (for example) `bash -c`.

`bash -c "echo \"123\""` is not the same as `bash -c "echo" "123"`.

SamirTalwar: Correct, I want the capability to provide commands + args directly, not just a shell command. There's no reason to invoke a shell for `echo 123`, after all, if we can represent it as `["echo", "123"]`.

daniel-chambers: I see. @SandeepSamba do you have the capacity to implement a third methodology here?

nullxone:
@SandeepSamba is this already implemented?

rfcs/0001-packaging.md
SandeepSamba: I would prefer it to be just a single string command rather than `string[]`. We have to invoke the commands via a shell because without a shell, all the fancy(coloured output, formatting,etc) logging that the plugins provide are lost. Even if we provide a string[], the CLI wil concat them with spaces and call it with shell. And with the native packaging [RFC](https://github.com/hasura/ndc-hub/pull/142/files#diff-164523172c282a097a4d5a1bcddcbe7b46cc13ac16674653a764ca87632f3b5bR48) we will anyway have to bring in a shell. @SamirTalwar Do you have any particular issue due to the choice of single string vs string[]?

SandeepSamba:
> @SandeepSamba is this already implemented?

No, this will have to to be done once the RFC is approved

rfcs/0001-packaging.md
codedmart: @SandeepSamba the problem is which shell? Isn't that the point of this. To not use a shell, and just have the CLI run the docker command. Bash isn't normally or always available on windows.

daniel-chambers: @codedmart I think @SandeepSamba's point is that he doesn't want to implement `string[]` (which bypasses the shell) as an alternative to `string` (which uses the shell), because bypassing the shell stuffs up formatting. The `DockerizedCommand` is separate to `string` and `string[]`. @SamirTalwar's suggestion of adding `string[]` was _in addition_ to `DockerizedCommand`.

SandeepSamba (approved):

PR #138, opened 2024-03-29
ndc-postgres v0.5.2

soupi:
https://github.com/hasura/ndc-postgres/releases/tag/v0.5.2

### DIFF
--- a/registry/aurora/metadata.json
+++ b/registry/aurora/metadata.json
@@ -7,7 +7,7 @@
     "tags": [
       "database"
     ],
-    "latest_version": "v0.5.1"
+    "latest_version": "v0.5.2"
   },
   "author": {
     "support_email": "support@hasura.io",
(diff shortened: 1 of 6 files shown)


registry/aurora/metadata.json
```diff
@@ -21,8 +21,8 @@
     "repository": "https://github.com/hasura/ndc-postgres",
     "version": [
       {
-        "tag": "v0.5.1",
-        "hash": "bdb2702",
+        "tag": "v0.5.2",
+        "hash": "805bc53",
```
scriptnull, line 25: Hi, do we want to get rid of v0.5.1 and have only v0.5.2 in our hub registry?

soupi: Good point. I think we can do that because version 0.5.1 haven't been used by users yet.

plcplc (approved):

PR #142, opened 2024-04-02
RFC: Native Connector Packaging

daniel-chambers:
[Current version - rendered](https://github.com/hasura/ndc-hub/blob/daniel/rfc-native-packaging/rfcs/0004-native-packaging.md)

[Original version - rendered](https://github.com/hasura/ndc-hub/blob/612956a842f0ffee1b88762f1f54f267973a26d1/rfcs/0004-native-packaging.md)


nullxone:
1. What if we did this instead

```typescript
type ConnectorMetadataDefinition = {
  packagingDefinition: { // keys represent the ConnectorManifest.definiton.type
    local?: PackagingDefinition;
    local-native?: PackagingDefinition;
  }
}

type PackagingDefinition = PrebuiltDockerImagePackaging | ManagedDockerBuildPackaging | ManagedLocalPackaging

type PrebuiltDockerImagePackaging { Commands, ... }
type ManagedDockerBuildPackaging { Commands, ... }
type ManagedLocalPackaging { Commands, ... }
```

It seems closer to the underlying model? Seemed odd to include supportsNativeToolchain inside ManagedDocker.

I recall that we want to encourage users to use the local-native version if they're developing that subgraph.

However, keeping the types closer to the underlying model allows us to explore those options easily in the future, wherein user only wants to iterative local-native on one connector in that subgraph.

2. What if the Commands.update and Commands.watch scripts themselves checked for dependencies on start.

This way:
- We don't need to worry about when to trigger/hook in this Command.init (i.e. we remove Command.init completely)
- We can support user flows where it will do the correct thing regardless of which ddn command the user uses as an entrypoint. For example, user clones repository, does ddn build connector-manifest (or ddn build supergraph-manifest) (or ddn update connector-link) (or ddn dev) etc. All of them just work.
- Commands.update and Commands.watch can specify a different set of dependencies if need be. For example, Commands.watch want some file monitoring program. Or maybe for some very large compiled connector, Commands.update is what we really want (??)

Maybe performance will be impacted? But I do not think so. Checking for dependencies should be super fast.

daniel-chambers:
@nullxone 

### Point 1

> What if we did this instead [...]

Hrm, that's a very breaking change to the connector-metadata.yaml format. The proposed change is trying to ensure that existing connectors would not be broken and the CLI can just incrementally adopt the additions, plus older CLI versions just ignore the new additions and retain the existing behaviour. 

Your proposed structure also lets you omit required `ConnectorManifest.definiton.type`s (realistically every connector would need to define at least `cloud` and one of `local` and `local-native`).

We _could_ make a breaking change by adding a `version: 2` to the top level and the CLI has to support both, but that wouldn't be compatible with old CLI versions. Maybe this is a non-issue?

> Seemed odd to include supportsNativeToolchain inside ManagedDocker.

The reason the proposed solution puts the native tooling settings underneath `ManagedDockerBuild` and not separately beside it, is because `ManagedDockerBuild` describes both how to build the connector for the cloud as well as locally in Docker. Any `ManagedDockerBuild` connector can be run locally as a container (as that was the original design), so the native tooling is really just an extension of that paradigm that describes how to additionally run the connector natively if we choose to. And for this first version we're going to _always_ choose to run it locally.

Also `PrebuiltDockerImage` connectors never need to specify local native tooling as they will always run locally using a Docker container.

I can see a good argument to making that `supportsNativeToolchain` property an object rather than a boolean to provide compatible space to, as you say, allow us to explore it more in the future.

### Point 2

This is a good point, we _could_ make `update`/`watch` scripts just restore packages. I like this idea. (For clarity, NodeJS Lambda does not have an `update`)

> Maybe performance will be impacted? But I do not think so. Checking for dependencies should be super fast.

This is not necessarily true, especially with `npm` (`npm` sucks). A clean package install (`npm ci`) always deletes `node_modules` and re-creates it, and while it does so from a cache in `~/.npm`, it's still not instant. A normal package install (`npm install`) _may_ be quicker if `node_modules` already exists since it doesn't delete and replace, but it also does extra work of comparing `package.json` and `package-lock.json` and maybe updating the latter from the former after a dependency re-compute.

The sample scripts in the RFC kinda hacked around this by checking if `node_modules` exists and not doing an `npm ci` if it does, but the more I think about this, the less I like it, because if the user gets the latest version of their project from git and if the deps have changed, running `ddn dev` won't "just work" because package restore will be skipped.

`yarn` does a better job of package management than `npm` (faster!) but the reason we're not using it is because we wanted to use the most "normal" and default NodeJS setup, and using `yarn` requires the user to enable NodeJS's corepack which is disabled by default (and sometimes not even installed depending on where they installed NodeJS from). It's messy, which is why we just went with `npm`. We could reassess this decision if the performance of using `npm` is truly not good enough, but IMHO we should really try not to.

daniel-chambers:
I'd also like to hear people's thoughts about the open questions and the alternative solution (CLI-embedded shell interpreter). 😄 

**Open Questions**
* What happens if the user runs `ddn build connector-manifest`? @nullxone This sends the connector to be built and deployed in the cloud, so this is not impacted by local native tooling, correct?
* Is the proposed solution workable for the Go functions connector? @hgiasac, any thoughts?
* Are there other things we'd rather do with an "init" command in the future? For example, delegate some of ddn quickstart process to the individual connectors so they can better validate user inputs such as connection strings etc? If so, how does that affect this? @nullxone Sounds like if we remove the `init` we can defer this one.

**Alternative solution (CLI-embedded shell interpreter)**: @nullxone I assume you agree that this is worse than having to provide two scripts (pwsh and bash)?

hgiasac:
@daniel-chambers This solution is workable for Go SDK. Shell scripts are also flexible for most of the use cases. Thanks for the RFC though

hgiasac:
About the `init` command, IMO we already package the connector boilerplate in the connector definition manifest, so it isn't required. However, it would be useful if the CLI automatically runs the `init` command when running the ` add connector-manifest` command

daniel-chambers:
@hgiasac Akshaya has suggested that we roll the `init` into `update` and `watch` (see his comments above). I believe `update` is called by `ddn add connector-manifest`, so whatever was going into `init` can go there instead; that sound okay to you?

daniel-chambers:
@nullxone Thinking about it some more, the way the proposed solution has things structured co-opts the `commands` structure if you `supportsNativeTooling`. This means that we kinda can't support the ability to have `local-native` and `local` `ConnectorManifest.definition.type`s at the same type, since `local` may need Dockerized/CLI plugin-ed commands and `local-dev` may need script commands. Was this what you were trying to point out?

If so, what if we did it like this instead:

```typescript
type ConnectorMetadataDefinition = {
  packagingDefinition: PackagingDefinition
  nativeToolchain?: NativeToolchainDefinition // **NEW**!
  supportedEnvironmentVariables: EnvironmentVariableDefinition[]
  commands: Commands
  cliPlugin?: CliPluginDefinition
  dockerComposeWatch: DockerComposeWatch
}

type NativeToolchainDefinition = {
  commands: Commands
}

type Commands = {
  // init removed
  update?: string | DockerizedCommand | ShellScriptCommand // *** Supports ShellScriptCommand now ***
  watch?: string | DockerizedCommand | ShellScriptCommand // *** Supports ShellScriptCommand now ***
}
```

We remove the `supportsNativeToolchain` from `ManagedDockerBuildPackaging` and add a top level `nativeToolchain` that contains a different set of commands that should be used when running with the native toolchain. It's separate from the packaging definition, because the argument can be made that `packagingDefinition` is about how to "package", and we're not packaging when using a native toolchain. However, every connector still needs a `packagingDefinition` because every connector still needs to be packaged when it is deployed on cloud.

Thoughts?

hgiasac:
> @hgiasac Akshaya has suggested that we roll the `init` into `update` and `watch` (see his comments above). I believe `update` is called by `ddn add connector-manifest`, so whatever was going into `init` can go there instead; that sound okay to you?

Yes, that also works

rfcs/0004-native-packaging.md
```diff
@@ -0,0 +1,173 @@
...
+type EnvironmentVariableDefinition = {
+  name: string
+  description: string
+  defaultValue?: string
+}
+
+type Commands = {
+  init?: string | DockerizedCommand | ShellScriptCommand // *** NEW! ***
+  update?: string | DockerizedCommand | ShellScriptCommand // *** Supports ShellScriptCommand now ***
+  watch?: string | DockerizedCommand | ShellScriptCommand // *** Supports ShellScriptCommand now ***
```
0x777, line 123: For completeness, do you also want to add a `run` to run a connector? Or is this only meant for local dev?

daniel-chambers: This is only meant for operations the CLI needs to perform with the connector, which in practice, means local dev. When running the connector "for realsies" (ie deployed in the cloud), we run using Docker, and how to do that is in the `packagingDefinition` (ie. `PrebuildDockerImage` or `ManagedDockerBuild`).

daniel-chambers:
@0x777 Afaik, Git for Windows bundles [msys2](https://www.msys2.org/) which contains a version of bash compiled for Windows, and a lot of the standard *nix tools like grep awk etc compiled for Windows. It's non-trivial and not small to ship something similar ourselves. I don't think I'd recommend it, but it could be possible. Maybe we could insist they have Git for Windows installed already and rely on the bash in there?

This is what my Git install looks like (notice the Linux-style root filesystem folders...):
image

daniel-chambers:
I've updated the RFC with all the feedback so far, after a meeting with @nullxone.

Please check over it one last time before we merge and commit to the design! 😄

nullxone:
LGTM!

rfcs/0004-native-packaging.md
```diff
@@ -0,0 +1,238 @@
...
+**connector-metadata.yaml**
+
+```yaml
+packagingDefinition:
+  type: ManagedDockerBuild
+nativeToolchainDefinition:
+  commands:
+    start:
+      type: ShellScript
+      bash: ./start.sh
```
SandeepSamba, line 48: Can we consider [multi-line](https://stackoverflow.com/a/21699210) yaml strings instead of filepaths to shell scripts (Its for the sake of implementation ease, nothing else)?

rfcs/0004-native-packaging.md
```diff
@@ -0,0 +1,238 @@
...
+```
+
+**.hasura-connector/start.sh**
+```bash
+#!/usr/bin/env bash
+set -eu -o pipefail
+
+./check-reqs.sh
+cd $HASURA_PLUGIN_CONNECTOR_CONTEXT_PATH
+exec npm run start
```
SandeepSamba, line 77: This script will be run as a sub-process by the CLI. Since this process will be in the background, How will the user access the logs of this process? With Docker, we could just (ask the user to) do `docker compose logs ...` with the service name. One solution is to pipe the connector logs to main CLI output but if there are multiple connectors running, it might become very noisy. Thoughts?

rfcs/0004-native-packaging.md
```diff
@@ -0,0 +1,238 @@
...
+```
```
SandeepSamba, line 78: Also, what is the restart policy for watch (or shoud we even have one?)

rfcs/0004-native-packaging.md
```diff
@@ -0,0 +1,238 @@
...
+  path: string
+  action: "rebuild" | "sync" | "sync+restart"
+  target?: string
+  ignore?: string[]
+}
+```
+
+## Alternative Solutions
+### Embedded Shell
+The most unappealing part of the proposed solution is the need to write two shell scripts if using `ShellScriptCommand`, in order to support both Linux/Mac (Bash) and Windows (PowerShell). We could try embedding a shell interpreter in the `ddn` CLI and using that to execute a shell script instead. For example, the [`interp`](https://pkg.go.dev/mvdan.cc/sh/v3@v3.8.0/interp) package purports to support a basic POSIX shell (with some Bash support).
```
SandeepSamba, line 227: Lets not do this please. Whenever I hear something `almost` works, its an obscure bug waiting to happen

rfcs/0004-native-packaging.md
daniel-chambers: The intention is that this is a bash command you can pass to the bash shell, not a filename. The command just happens to be invoking a script file here. I'll clarify this in the spec.

We definitely want the potential for script files to be involved since there'll likely be some shared logic between commands.

rfcs/0004-native-packaging.md
daniel-chambers: Good question. 🤔 You could redirect stderr/stdout to a temp file instead and tell the user to tail (bash)/Get-Content -Tail n (powershell) the file?

rfcs/0004-native-packaging.md
daniel-chambers: Personally, longer term, I'd like to see `ddn dev` accept user input while watching to allow manual restart of misbehaving watchers. Building on that idea, perhaps if a watch fails, the user could press a key to restart it?

rfcs/0004-native-packaging.md
daniel-chambers: I agree, it's risky af. :)

codedmart (approved):

PR #154, opened 2024-04-16
update ndc-postgres to v0.6.0

soupi:
https://github.com/hasura/ndc-postgres/releases/tag/v0.6.0

### DIFF
--- a/scripts/update-ndc-postgres.sh
+++ b/scripts/update-ndc-postgres.sh
@@ -1,4 +1,8 @@
-#!/bin/bash
+#!/usr/bin/env bash
+
+set -e
+set -u
+set -o pipefail
 
 # Update ndc-postgres related entries with a new version.
 
--- a/scripts/update-ndc-postgres.sh
+++ b/scripts/update-ndc-postgres.sh
@@ -29,18 +29,20 @@ for variant in \
     'postgres-timescaledb' ; do
 
 # add new version
-cat registry/"${variant}"/metadata.json | jq \
+jq \
     --arg tag "${TAG}" \
     --arg hash "${HASH}" \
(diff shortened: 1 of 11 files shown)


scripts/update-ndc-postgres.sh
soupi, line 1: Added a script to make this update easier in the future.

scripts/update-ndc-postgres.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
```
SamirTalwar, line 2: I suggest failing fast.

```suggestion

set -e
set -u
set -o pipefail

```

scripts/update-ndc-postgres.sh
```diff
@@ -0,0 +1,44 @@
...
+    'neon' \
+    'postgres' \
+    'postgres-alloydb' \
+    'postgres-azure' \
+    'postgres-cosmos' \
+    'postgres-gcp' \
+    'postgres-timescaledb' ; do
+
+# add new version
+cat registry/"${variant}"/metadata.json | jq \
```
SamirTalwar, line 28: `cat $FILE | jq $QUERY` can be simplified to `jq $QUERY $FILE`.

scripts/update-ndc-postgres.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
```
SamirTalwar, line 1: Bash doesn't always live in `/bin`.

```suggestion
#!/usr/bin/env bash
```

SamirTalwar (approved):

soupi:
Thanks for the review @SamirTalwar !

PR #164, opened 2024-05-09
Docker CLI plugin RFC

daniel-chambers:
[Rendered](https://github.com/hasura/ndc-hub/blob/daniel/rfc-docker-cli-plugin/rfcs/0005-docker-cli-plugin.md)


SandeepSamba (approved):
LGTM from `ddn` CLI end. It would be great if we can get this validated from the connector author end as well.

PR #204, opened 2024-07-18
Specify ndc-models version for each postgres release

danieljharvey:
We'd like to include the `ndc_models` version requested by a user as an env var, so we are using this as a mapping between desired connector version and underlying `ndc_models` version.

A few of the Postgres variants were out of line, I just fixed one and copy pasted so they should all match now.


scriptnull (commented):
cc: @codingkarthik (for the awareness of the addition of this new field)

PR #205, opened 2024-07-18
RFC: Env var mapping

SandeepSamba:
[Rendered](https://github.com/hasura/ndc-hub/blob/19ee2f4d5e48bea31befbe2a99410caf4b9bf50f/rfcs/0006-env-mapping.md)

### DIFF
--- a/rfcs/0006-env-mapping.md
+++ b/rfcs/0006-env-mapping.md
@@ -13,7 +13,7 @@ type EnvJSON = {
 type EnvMap = {
     [key: string]: Val
 }
-type Val = string | ValFromEnv
+type Val = { value: string } | ValFromEnv
 type ValFromEnv = {
     valueFromEnv: string
 }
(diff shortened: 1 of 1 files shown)


nullxone:
```typescript
// Instead of
type Val = string | ValFromEnv

// Let's do
type Val = { value: string } | ValFromEnv
```

It's easier for tooling to parse and handle this type. And this is ok because the user is not generally editing this file, it is autogenerated.

SandeepSamba:
> ```ts
> // Instead of
> type Val = string | ValFromEnv
> 
> // Let's do
> type Val = { value: string } | ValFromEnv
> ```
> 
> It's easier for tooling to parse and handle this type. And this is ok because the user is not generally editing this file, it is autogenerated.

Done

nullxone:
A solution that does not require changes to the connector spec is to define the connector.yaml as

```yaml
  kind: Connector
  definition:
    name: mypg
    hasura-hub: hasura/postgres:2
    envFile: ./.env.cloud2
    env:
      - CONNECTION_URI=${APP_MYPG_CONNECTION_URI}
```

Then we can still use a single env file, but do the mapping at the CLI level.

native (non docker) connectors still need some mapping, but we can assume that they mostly take user specified env vars (like with typescript), so we can solve that mapping later.

SandeepSamba:
Withdrawing the RFC as we will handle the mapping at the CLI end, similar to Akshaya's proposal.

PR #207, opened 2024-07-19
Release ndc-graphql v0.1.1

BenoitRanque:


codedmart (approved):
Approving, but don't we also want to update "latest_version"?

PR #212, opened 2024-07-29
RFC: Connector plugin to output NDC schema 

varundey:
[Rendered](https://github.com/hasura/ndc-hub/blob/print-schema-capabilities-rfc/rfcs/006-print-schema-capabilities.md)


rfcs/006-print-schema-capabilities.md
```diff
@@ -0,0 +1,100 @@
...
+type EnvironmentVariableDefinition = {
+  name: string
+  description: string
+  defaultValue?: string
+}
+
+type Commands = {
+  update?: string | DockerizedCommand | ShellScriptCommand 
+  watch?: string | DockerizedCommand | ShellScriptCommand
++ printSchema?: string |  DockerizedCommand | ShellScriptCommand
```
daniel-chambers, line 71: @SandeepSamba Do we actually support `ShellScriptCommand` in the CLI? We'd need it for the NodeJS Lambda connector to implement this in a decent way.

By "decent", I mean with good error messages that check for native tooling and check that the user has restored packages (if not, show a good error).

rfcs/006-print-schema-capabilities.md
```diff
@@ -0,0 +1,100 @@
...
+
+This RFC proposes to add a new optional command `printSchema` to the connector plugin which prints out NDC schema of the underlying data source (the type [VersionedSchemaAndCapabilities](https://github.com/hasura/v3-engine-multitenant/blob/c6da0685556f8f31194a7f6620513ff5c7261497/v3-metadata-build-service/jsonschemas/metadata.jsonschema#L5636)) to STDOUT. CLI can then call the plugin with relevant command to get the schema and construct DataConnectorLink from the output. All connector plugins are expected to have this command going forward. DDN CLI will use this to generate NDC schema if this command is present. In case, the command is not implemented on plugin (in older versions of connectors), CLI will rely on running the connector in docker to get the schema.
```
daniel-chambers, line 5: `VersionedSchemaAndCapabilities` is an OpenDD type. OpenDD is not available in connectors; there is no dependency on OpenDD in connectors architecturally. 

I'd suggest adding a type definition to the SDK that performs a similar function to this type but is not coupled to OpenDD, such as:

```
type SchemaAndCapabilities = {
  schema: ndc.SchemaResponse,
  capabilities: ndc.CapabilitiesResponse
}
```

It'd be the CLI's responsibility to map this type into the OpenDD type.

soupi: Alternatively, we could introduce multiple commands, one for each ndc info component such as capabilities, schema, etc. Or we could introduce a composite type in ndc-spec for all of the information if it doesn't seem useful to have them separately.

SamirTalwar, line 5: Is it "optional" if "all connector plugins are expected to have this command going forward"?

rfcs/006-print-schema-capabilities.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Changes to the packaging spec
+There is a new command `print-schema` in the `commands` section: 
+
+```diff
+packagingDefinition:
+  type: ManagedDockerBuild/PrebuiltDockerImage
+supportedEnvironmentVariables: [...]
+commands:
+  update: hasura-ndc-postgres update
```
SamirTalwar, line 15:
```suggestion
  update: hasura-ndc-plugin update
```

rfcs/006-print-schema-capabilities.md
```diff
@@ -0,0 +1,100 @@
...
+}
+
+
+type DockerizedCommand = {
+  type: "Dockerized"
+  dockerImage: string 
+  commandArgs: string[]
+}
+
+type ShellScriptCommand = {
```
SamirTalwar, line 81: I am once again asking for a "plain" command which does not rely on a shell and is instead a list of arguments.

scriptnull:
I have a doubt: If this RFC is accepted and implemented, then the CLI does not need the `schema` field in the connector APIs for the connectors deployed in the cloud - is it a safe assumption to say?

rfcs/006-print-schema-capabilities.md
varundey: @daniel-chambers right now CLI only works with `string` type. Do you need this change for this RFC?

rfcs/006-print-schema-capabilities.md
varundey: updated 👍

rfcs/006-print-schema-capabilities.md
varundey: removed "optional". Thanks!

rfcs/006-print-schema-capabilities.md
varundey: Can we take this up in a separate RFC pls. Since this is already in place

rfcs/006-print-schema-capabilities.md
varundey: @daniel-chambers CLI currently only supports `string` type. Would you need `ShellScriptCommand` support for this RFC?

rfcs/006-print-schema-capabilities.md
varundey: CLI anyways calls the two separately and then maps them together right now. Please suggest what would be the best and fastest way forward.

varundey:
> I have a doubt: If this RFC is accepted and implemented, then the CLI does not need the `schema` field in the connector APIs for the connectors deployed in the cloud - is it a safe assumption to say?

No. That would still be relevant to fetch the schema for connectors deployed on cloud. This RFC is mostly around local development

rfcs/006-print-schema-capabilities.md
daniel-chambers: I would go with just adding the type to the SDK. Given we always want to refresh the capabilities and schema together it seems pointless to execute the program twice. But honestly @soupi's way is fine too. I don't really mind.

rfcs/006-print-schema-capabilities.md
daniel-chambers: Yes, for the reasons I stated above. Without it you'll get trash errors if the user doesn't have things set up properly.

varundey: I tested `ShellScriptCommand` by putting a dummy echo for `Command.update` and tried introspecting nodejs connector and it printed out the echo. So technically we do support `ShellScriptCommand`

daniel-chambers: > The pain point of ShellScriptCommand is cross-platform support, especially Windows PowerShell. I still prefer native CLI and Docker plugins for complicated scripts.

That's fine when you have something that is easily cross platform or entirely dockerized, but in some scenarios it's not that simple (in particular NodeJS Lambda Connector). I'm certainly not suggesting _removing_ support for doing it via Docker or CLI plugin command. So if its easier for you to use a CLI plugin command, definitely do that. I just want the `ShellScriptCommand` support available for connectors where that's the most pragmatic way of dealing with the problem.

> I tested ShellScriptCommand by putting a dummy echo for Command.update and tried introspecting nodejs connector and it printed out the echo. So technically we do support ShellScriptCommand

Nice! Just to clarify, did you test it on Windows with a PowerShell script?

varundey: Tried this on mac with shell. Could you please try it on windows powershell if you have access to one? I can help in testing it out if you need

daniel-chambers: @varundey It appears to work with PowerShell on Windows. Great!

rfcs/006-print-schema-capabilities.md
daniel-chambers: @varundey Would you prefer two commands, one for schema, the other for capabilities, to match your current behaviour?

daniel-chambers: An SDK change is probably how we'll implement it in either case. So connectors will update to the latest SDK and get the two new commands for "free".

daniel-chambers: And yes, @varundey, `printSchema` and `printCapabilities` (note casing to be consistent with everything else) is fine.

varundey: > An SDK change is probably how we'll implement it in either case. So connectors will update to the latest SDK and get the two new commands for "free".

Oh I meant updating the schema to have the composite types of schema and capabilities as a single command for plugin. That will take some time to get RFC'd and implement I am assuming.

rfcs/006-print-schema-capabilities.md
```diff
@@ -0,0 +1,122 @@
...
+## Changes to the packaging spec
+There is a new command `printSchemaAndCapabilities` in the `commands` section: 
+
+```diff
+packagingDefinition:
+  type: ManagedDockerBuild/PrebuiltDockerImage
+supportedEnvironmentVariables: [...]
+commands:
+  update: hasura-ndc-plugin update
++ printSchemaAndCapabilities: hasura-ndc-plugin print-schema-and-update
```
daniel-chambers, line 16: `print-schema-and-update`?

```suggestion
+ printSchemaAndCapabilities: hasura-ndc-plugin print-schema-and-capabilities
```

varundey: typo. have updated now

rfcs/006-print-schema-capabilities.md
daniel-chambers, line 1: Sorry to be a pain but could you rename this file to `0008-print-schema-capabilities.md` so the numbering lines up with what's already there?

varundey: thanks for the catch. done. 👍

daniel-chambers (approved):
This looks good to me now. 😄

PR #238, opened 2024-08-13
Update turso connector to v0.0.32

TristenHarr:
This PR updates the turso connector metadata to version 0.0.32.

### DIFF
--- a/registry/hasura/turso/metadata.json
+++ b/registry/hasura/turso/metadata.json
@@ -7,7 +7,7 @@
     "tags": [
       "database"
     ],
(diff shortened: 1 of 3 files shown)


registry/hasura/turso/metadata.json
```diff
@@ -16,19 +16,6 @@
   },
   "is_verified": false,
   "is_hosted_by_hasura": true,
-  "packages": [
```
gneeri, line 19: I'm not sure if we should be pulling this array out just yet. @codingkarthik could you weigh in?

PR #254, opened 2024-08-22
Enable the registry automation in prod

codingkarthik:
This PR copies the `registry-automation` GitHub workflow (that deploys to staging) to `registry-automation-prod` . This workflow will only run when there is a `push` to the `main` branch.


.github/workflows/registry-updates.yaml
```diff
@@ -1,17 +1,16 @@
...
       - main
-    types: [opened, synchronize, reopened]
     paths:
       - registry/**
 
 jobs:
   update_registry_db:
     runs-on: ubuntu-latest
-    environment: staging
+    environment: production
```
scriptnull, line 13: Did we make this change to test prod before merging the PR?

codingkarthik: Yes, I tested [here](https://github.com/hasura/ndc-hub/pull/255)! It is working correctly.

.github/workflows/registry-updates.yaml
```diff
@@ -1,17 +1,16 @@
-name: Update Hub DB from GH Registry
+name: Update Hub DB from GH Registry (Staging)
 
 on:
   pull_request:
     branches:
       - main
-    types: [opened, synchronize, reopened]
```
scriptnull, line 7: Any particular reason for removing this? Removing this will lead to triggering the action for a lot more events: https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request, right?

.github/workflows/registry-updates.yaml
codingkarthik, line 1: @scriptnull Sorry, some changes you saw before were not supposed to be committed, I was trying to make a single workflow for both the prod and staging, but it is not straight forward. Now, the prod workflow is in the `registry-updates-prod.yaml` file only.

.github/workflows/registry-updates-prod.yaml
```diff
@@ -0,0 +1,60 @@
+name: Update Hub DB from GH Registry (Staging)
```
scriptnull, line 1: nit: `(Staging)` -> `(Prod)`

.github/workflows/registry-updates-prod.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+      - name: Setup Go
+        uses: actions/setup-go@v4
+        with:
+          go-version: 1.21.x
+
+      - name: Run registry automation program
+        env:
+          CHANGED_FILES_PATH: "changed_files.json"
+          PUBLICATION_ENV: "staging"
```
scriptnull, line 51: Should we use Production? I wonder if we are making use of this env var :thinking:

codingkarthik: Yeah, we are not using the env var yet. I will change it to `production`.

scriptnull (approved):

PR #261, opened 2024-08-22
Update turso connector to v0.1.0

TristenHarr:
This PR updates the turso connector metadata to version 0.1.0.

### DIFF
--- a/registry/hasura/turso/metadata.json
+++ b/registry/hasura/turso/metadata.json
@@ -7,7 +7,7 @@
     "tags": [
       "database"
     ],
-    "latest_version": "v0.0.15"
+    "latest_version": "v0.1.0"
   },
   "author": {
     "support_email": "support@hasura.io",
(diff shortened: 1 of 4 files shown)


registry/hasura/turso/metadata.json
```diff
@@ -16,19 +16,6 @@
...
-      "version": "0.0.15",
-      "uri": "https://github.com/hasura/ndc-turso/releases/download/v0.0.15/connector-definition.tgz",
-      "checksum": {
-        "type": "sha256",
-        "value": "a2c8426c443bf8c94b0f706d8438bd06eeeef30c0d2b40389c0e37be4be920a6"
-      },
-      "source": {
-        "hash": "6bae967daf7b9aa718d62103e077a739c41d5a6d"
-      }
-    }
```
daniel-chambers, line 30: Why did this old version get deleted and not migrated to the releases folder?

TristenHarr: That version is already in the CLI, so there was no need to have it added to the CLI via the automation. 

This PR was opened by a [github workflow](https://github.com/hasura/ndc-turso/blob/main/.github/workflows/docker.yml). The workflow automates the opening of a PR into ndc-hub and updating the README etc.

If preferred I can go in and create the releases folder by hand.

daniel-chambers: Yeah except this repo is supposed to contain what the available versions are, and deleting it here means that version only magically exists in some backend CLI service that nobody can see. This repo should continue to properly represent an inventory of all versions.

daniel-chambers (approved):

PR #257, opened 2024-08-22
Add SingleStore connector

AdalbertMemSQL:
This PR adds singlestore connector.

PR #268, opened 2024-08-29
Allow forked PRs to deploy connector version publishing in the staging environment

codingkarthik:
This PR extends the `registry-updates` workflow to access the repository secrets for forked repository PRs, like this [one](https://github.com/hasura/ndc-hub/pull/257) . 

It should be noted that, we are allowing this only for the staging environment and this is not needed for the production environment.

### DIFF
--- a/.github/workflows/registry-updates.yaml
+++ b/.github/workflows/registry-updates.yaml
@@ -19,6 +19,26 @@ jobs:
         with:
           fetch_depth: 1
 
+
+      - name: Check for PR approvals
+        id: check-approval
+        uses: actions/github-script@v6
+        with:
+          script: |
+            const { data: reviews } = await github.pulls.listReviews({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.payload.pull_request.number,
+            });
+            const approved = reviews.some(review => review.state === 'APPROVED');
+            return approved;
+
+      - name: Cancel if not approved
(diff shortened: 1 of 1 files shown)


imperfect-fourth (commented):
i don't like the idea of PRs being able to update our DB(even if it's staging only). Can we instead update on merge, or when PR is approved?

codingkarthik:
> Can we instead update on merge, or when PR is approved?

Yes, I can change the workflow to run after the PR is approved. Basically, the idea initially was for the connector authors to re-iterate on the PR until they are satisfied with the changes, but I don't think it happens a lot, so we can totally make this change.

.github/workflows/registry-updates.yaml
```diff
@@ -19,6 +19,26 @@ jobs:
...
+              pull_number: context.payload.pull_request.number,
+            });
+            const approved = reviews.some(review => review.state === 'APPROVED');
+            return approved;
+
+      - name: Cancel if not approved
+        if: steps.check-approval.outputs.approved == 'false'
+        run: |
+          echo "The PR has not been approved. Cancelling the workflow."
+          exit 1  # This will exit the workflow early
```
imperfect-fourth, line 40: will this show the workflow run as failed?

codingkarthik: Yes, until the PR is not approved the workflow will be shown as failed.

imperfect-fourth (approved):

PR #277, opened 2024-09-02
release openapi connector version `v0.1.2`

m-Bilal:


registry/hasura/openapi/metadata.json
```diff
@@ -58,6 +58,17 @@
       "source": {
         "hash": "4b70712933de7a7ed475407ccd0be7bed1611ae8"
       }
+    },
+    {
```
daniel-chambers, line 62: The packages section is no longer required now that you've got a releases folder. That folder replaces this.

m-Bilal: Understood. I assumed it had something to do with compatibility for publishing the connector. Resolved in https://github.com/hasura/ndc-hub/pull/277/commits/7ba13fc6c9ca510107ad1b9c8d5136d5c9d5d04b

daniel-chambers (approved):
@m-Bilal I'd probably delete the whole packages property from the metadata.json, since it's replaced with the releases folder. Just so things aren't inconsistent and there's a single source of truth. 

This will work though, since nothing actually looks at packages.

PR #296, opened 2024-09-23
Connector upgrades DX

nullxone:
Introduces a new plugin command `upgradeConfiguration` to upgrade connector configuration in place.

[Rendered](https://github.com/hasura/ndc-hub/blob/ak/connector-upgrade-dx/rfcs/0010-connector-upgrades-dx.md)


varundey (commented):
fwiw CLI can already upgrade the plugin using `ddn plugin upgrade <plugin> --version <target-version>`. Unsure if connector author should implement this chage

rfcs/0009-connector-upgrades-dx.md
```diff
@@ -0,0 +1,98 @@
...
+
+How does a user upgrade across connector versions? For example: from postgres v0.7.0 to v1.1.1 ?
+
+## Solution
+
+`ddn connector upgrade --connector <path-to-connector.yaml> [--version <hub-version>]` will
+- upgrade the connector to the specified version or to the latest version
+- upgrade the connector.yaml and compose.yaml files in place
+- atomically replace the .hasura-connector folder
+- call a new plugin command `upgradeConfiguration`
```
daniel-chambers, line 11: I assume the plugin from the new version of the connector is called? We should make that explicit.

rfcs/0009-connector-upgrades-dx.md
```diff
@@ -0,0 +1,98 @@
+# RFC: connector upgrades DX
+
+How does a user upgrade across connector versions? For example: from postgres v0.7.0 to v1.1.1 ?
+
+## Solution
+
+`ddn connector upgrade --connector <path-to-connector.yaml> [--version <hub-version>]` will
+- upgrade the connector to the specified version or to the latest version
+- upgrade the connector.yaml and compose.yaml files in place
+- atomically replace the .hasura-connector folder
```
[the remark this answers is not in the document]
nullxone: CLI should continue to install plugins as it currently does.

Only the .hasura-connector folder that is a part of the user project is to be replaced.

nullxone:
> fwiw CLI can already upgrade the plugin using `ddn plugin upgrade <plugin> --version <target-version>`. We can maybe extend the functionality to be called from `connector` cmd. Unsure if connector author should implement this chage

CLI will upgrade the plugin, compose.yaml, connector.yaml .

Connector plugin command `upgradeConfiguration` should upgrade **only** the files owned by the connector. Like for example configuration.json for postgres, or package.json for typescript.

rfcs/0009-connector-upgrades-dx.md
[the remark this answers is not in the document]
varundey: ah my bad. I assumed it is the `.hasura-connector` in `.ddn` dir

rfcs/0010-connector-upgrades-dx.md
```diff
@@ -0,0 +1,104 @@
...
+- upgrade the connector to the specified version or to the latest version
+- upgrade the connector.yaml and compose.yaml files in place
+- atomically replace the .hasura-connector folder
+- call the plugin command `upgradeConfiguration` on the *target version* of the connector
+
+The target version must be greater than the current version.
+
+The plugin command is expected to perform any connector specific upgrades to the connector configuration.
+
+If the plugin command is not defined, the CLI will WARN that the connector does not natively support configuration upgrades and that manual upgrades may be necessary.
```
m-Bilal, line 19: Does the CLI warn before performing the upgrade or after? I feel we should warn users before performing the upgrade and users should then opt in fully aware that there will manual work required. I say this because `configuration.json` files (and other files owned by connectors) are not documented anywhere, and performing manual upgrades may not be straightforward.

nullxone: This is a fair idea, we can play around with this on the implementation side depending on how it lands with users.

Will fold this into the CLI, but considering this out of scope for this RFC.

daniel-chambers (approved):

PR #303, opened 2024-09-25
Add ndc-prometheus v0.0.4

hgiasac:


scriptnull (approved):
We currently have the automation for publishing new connector versions, but when we introduce a new connector, we might need to create some entries manually for now. @codingkarthik is working on automating that part too. So if the github action fails even after approval, let us know and one of us can create the entries for you!

PR #322, opened 2024-10-03
RFC: Include CLI packaging information within the connector metadata 

codingkarthik:
# Add Binary CLI Plugin Packaging Information to `connector-metadata.yaml`

This PR introduces changes to improve the management of binary CLI plugin information in the `connector-metadata.yaml` file.

## Key Changes

1. Extended `connector-metadata.json` to include binary CLI plugin packaging information.
2. Added a new optional `version` field to the `ConnectorMetadataDefinition` type for future versioning.
3. Extend `BinaryCliPluginDefinition` to accommodate inline binary CLI definition.

## Addition of `connector-metadata-types`

A significant part of this PR is the introduction of `connector-metadata-types`. This file contains TypeScript type definitions for the `connector-metadata.json` structure and is meant to act as the source of truth for the schema of the `connector-metadata.json` file. As we advance, it is expected that if any changes are proposed in the connector metadata structure, they also update the `connectorTypes.ts` file in the `connector-metadata-types` folder.


rfcs/0011-cli-and-connector-packaging.md
```diff
@@ -0,0 +1,154 @@
...
+### `connector-metadata.yaml` types
+
+```typescript
+export type ConnectorMetadataDefinition = {
+  version?: 1;
+  packagingDefinition: PackagingDefinition;
+  nativeToolchainDefinition?: NativeToolchainDefinition;
+  supportedEnvironmentVariables: EnvironmentVariableDefinition[];
+  commands: Commands;
+  cliPlugin?: BinaryCliPluginDefinition | DockerCliPluginDefinition;
```
daniel-chambers, line 35: This new type for `cliPlugin` is a breaking change to the metadata format. We should avoid breaking changes. Can this be expressed in a non-breaking way?

codingkarthik: It is not actually a new type, because this is the [schema](https://github.com/hasura/v3-cli-go/blob/main/pkg/types/connector/pkgdef/types.go) that's being used and in that, the type of `cliPlugin` is the same. The introduction of this type was in this [RFC](https://github.com/hasura/ndc-hub/blob/main/rfcs/0005-docker-cli-plugin.md), but unfortunately it didn't carry forward in the subsequent RFCs.

connector-metadata-types/tsconfig.json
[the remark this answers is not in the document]
codingkarthik: Oh, I didn't know about this. Thanks!

rfcs/0011-cli-and-connector-packaging.md
```diff
@@ -0,0 +1,154 @@
...
+  /**
+   * The SHA256 hash of the binary file. This is used to verify the integrity of the downloaded binary.
+   */
+  sha256: string;
+  /**
+   * The name of the binary file. This is the name of the binary file that will be placed in the bin directory.
+   */
+  bin: string;
+
+  files: File[];
```
varundey, line 125: Think this meant to be `FilePath`?

rfcs/0011-cli-and-connector-packaging.md
```diff
@@ -0,0 +1,154 @@
...
+};
+
+export type FilePath = {
+  from: string;
+  to: string;
+};
+
+export type BinaryCliPluginDefinition = {
+  type: "Binary";
+  name: string;
```
varundey, line 135: should have a version field as well

rfcs/0011-cli-and-connector-packaging.md
codingkarthik: Yes, I have updated it!

rfcs/0011-cli-and-connector-packaging.md
codingkarthik: Added it, thanks for pointing this out.

rfcs/0011-cli-and-connector-packaging.md
daniel-chambers: Yes, but unfortunately it's still a breaking change to that type because:

* `BinaryCliPluginDefinition` is missing a version field (`version: string`)
* `BinaryCliPluginDefinition`'s type property should be optional (ie `type?: "Binary"`) (because before this was a union, there was no type property)

Also, what does it mean for a `BinaryCliPluginDefinition` to have a `version` and also `platforms`? Perhaps it would be better to separate the old `BinaryCliPluginDefinition` (which means "get it from the cli plugin index") from the new definition by creating a new type `BinaryInlineCliPluginDefinition` which has `platforms` but no `version` (and `type: "BinaryInline") so that way its very clear that one can either have a `version` or `platforms` but not both.

connector-metadata-types/connectorTypes.ts
daniel-chambers, line 1: Why is there both `connector-metadata-types/connectorTypes.ts` and `connector-metadata-types/src/connectorTypes.ts`? They seem to be the same...

rfcs/0011-cli-and-connector-packaging.md
daniel-chambers: That sounds good @codingkarthik. Keep in mind that we'll need to support `BinaryCliPluginDefinition` for quite some time until the connector versions that use it (ie. all of them right now) are no longer used by users.

connector-metadata-types/src/connectorTypes.ts
```diff
@@ -0,0 +1,135 @@
...
+  /**
+   * The SHA256 hash of the binary file. This is used to verify the integrity of the downloaded binary.
+   */
+  sha256: string;
+  /**
+   * The name of the binary file. This is the name of the binary file that will be placed in the bin directory.
+   */
+  bin: string;
+
+  files: FilePath[];
```
daniel-chambers, line 104: What is `files` for and how does it differ/relate to `bin`, `uri`, `sha256`? Might deserve some comments around this and the `FilePath` type.

I know this is basically a copy/paste of what's in the plugin index, but it seems weird to me to have an array of files here but really only support one file, since there's only one `uri`, `bin` and `sha256`...

connector-metadata-types/src/connectorTypes.ts
```diff
@@ -0,0 +1,135 @@
...
+  from: string;
+  to: string;
+};
+
+export type BinaryCliPluginDefinition =
+  | BinaryExternalCliPluginDefinition
+  | BinaryInlineCliPluginDefinition;
+
+export type BinaryInlineCliPluginDefinition = {
+  type?: "BinaryInline";
```
daniel-chambers, line 117: This should not be an optional property (ie `type?`). Optionality is only required for `BinaryExternalCliPluginDefinition` because it needs to support old versions. It should be required here.

rfcs/0011-cli-and-connector-packaging.md
codingkarthik: @SandeepSamba As per our discussion, the `name` and the `version` of the inlined binary CLI plugin definition will be the same as the connector name and version, and in case the connector wants to use a different version of the Binary CLI plugin, they will have to specify the `name` and the `version` of the Binary CLI plugin they want to use.
cc: @daniel-chambers @scriptnull

connector-metadata-types/src/connectorTypes.ts
codingkarthik: Makes sense, Thanks for pointing it out!

connector-metadata-types/src/connectorTypes.ts
codingkarthik: @SandeepSamba Could you please help out here? I'm not sure what the `files` field meant in the CLI plugin manifest which is why I left it empty here.

varundey: [This](https://github.com/hasura/cli-plugins-index/blob/master/plugins/ndc-postgres/v1.1.1/manifest.yaml#L10-L12) is an example of your `BinaryCliPluginPlatform` type. `files` from and to is basically the name of the binary that is going to get downloaded from the url and what it should be renamed to. CLI does all of this right now and there probably is a good scope to loosen this weird interface

daniel-chambers: Yeah, so why does it need to be a list? Feels like the whole `files` thing is redundant, you could just rename the file downloaded from `uri` to the filename in `bin`. Am I missing something?

varundey: I agree. I don't know the use case it was introduced for but as I said we can greatly improve this entire thing.

codingkarthik: Folks, then I will go ahead and remove the `files` thing.

SandeepSamba: @codingkarthik `files` is needed if we have to ship some supporting files to the plugin (For eg. We need to provide some `.dll` files for the windows binary to work correctly, supporting header files for a plugin which is not statically linked. Also think of a use case where the `bin` file (the file that is invoked by the CLI) is a `.bat` or a `.sh` file and it calls multiple binaries inside them). Without `files` then all we can do is to unzip the entire archive and place everything as it was in the archive in the plugins directory. `from` and `to` is to support renaming when the plugin is being unpacked and installed (Can also be used to change the location of the file via renaming). We can remove the `files` field, but then we lose out on these flexibilities. cc @daniel-chambers

daniel-chambers: Honestly, if it were me, I'd just require every plugin to be a .tar.gz (or zip, or whatever archive format you want) and require connectors to publish their files with the correct filenames inside that archive. Why have the complexity of having a custom file renaming system (`files`) and requiring connectors to define the renames here and the CLI to implement understanding it and do file renaming? We have technology already that represent files with names potentially in directories, wrapped up in a single file: tar/zip files. We should just use it.

SandeepSamba (approved):

daniel-chambers (approved):

PR #328, opened 2024-10-07
fix sha256 checksum for some connectors

scriptnull:
Noticed these inconsistencies while working on https://github.com/hasura/ddn-assets/pull/4


codingkarthik (approved):

hgiasac (approved):

codingkarthik:
@scriptnull I think it would be a good idea to validate the checksum value during the connector publishing automation. WDYT?

scriptnull:
> I think it would be a good idea to validate the checksum value during the connector publishing automation.

Yep, I think we will eventually arrive at that state.

PR #353, opened 2024-10-18
Adds feature tables to each of the connectors on the hub

BrittonHasura:


typhonius:
@codedmart this takes all the feature tables on individual ndc-* connectors and updates the connector hub.

PR #383, opened 2024-11-13
Record latest version, title and description if the connector metadata changes

codingkarthik:
Every time the connector's metadata.json file changes, the `latest_version`, `title` and `description` will be updated to the latest values in the registry DB.


scriptnull (approved):
Approving to unblock testing the CI run. Will take a closer look when I get the chance.

PR #460, opened 2025-02-07
validate the latest version when new connector version is released

codingkarthik:
Same as title


m-Bilal (approved):
rubberstamp

.github/workflows/registry-updates.yaml
```diff
@@ -23,7 +23,6 @@ jobs:
           repository: ${{github.event.pull_request.head.repo.full_name}}
           fetch-depth: 1
 
-      - name: Check for PR approvals
```
scriptnull, line 26: Removed by mistake? :thinking:

registry-automation/cmd/ci.go
```diff
@@ -245,6 +250,61 @@ func processChangedFiles(changedFiles ChangedFiles) ProcessedChangedFiles {
 	return result
 }
 
+// validateLatestVersion checks if the latest version in metadata.json matches
+// the actual latest version from the releases directory
+func validateLatestVersion(connector Connector, declaredLatestVersion string) error {
```
scriptnull, line 255: The logic of this function looks good to me :+1: 

Could we move this function to the `validate` package [here](https://github.com/hasura/ndc-hub/tree/main/registry-automation/pkg/validate)?

Rationale behind it:
- We introduced a new command called "validate" [here](https://github.com/hasura/ndc-hub/blob/main/registry-automation/cmd/validate.go).
- The validate command gets run on all commits (even before someone from Hasura adds approval to a PR) via [this GitHub action](https://github.com/hasura/ndc-hub/blob/main/.github/workflows/validate.yaml). That way a connector author can get immediate feedback and fix all the validation errors even before someone from Hasura takes a look at this PR for approval.
- I see that you would like to run the validations before we go for an update in the DB etc. Could we run the validate command before invoking the `ci` command to get all the validations? (or perhaps internally invoke `validate` command as the first thing in the `ci` command) - Trying to arrive at some way where we can run the validation separately and also along with the updates.

.github/workflows/registry-updates.yaml
codingkarthik: Yup, will revert it back. Thanks!

registry-automation/cmd/ci.go
codingkarthik: Yes, I will move the function to the `validate` package and I think it should be fine only to run the validations once every commit and not worry about running the validations with the updates.

registry-automation/cmd/validate.go
```diff
@@ -62,6 +72,32 @@ func executeValidateCmd(cmd *cobra.Command, args []string) {
 			}
 		}
 
+		// Check for metadata.json files
+		if filepath.Base(path) == "metadata.json" && !strings.Contains(path, "aliased_connectors") {
```
scriptnull, line 76: we can use `ndchub.MetadataJSON` instead of `"metadata.json"`

registry-automation/cmd/validate.go
```diff
@@ -62,6 +72,32 @@ func executeValidateCmd(cmd *cobra.Command, args []string) {
...
+			metadata, err := os.ReadFile(path)
+			var connectorMetadata ConnectorMetadata
+			err = json.Unmarshal(metadata, &connectorMetadata)
+			if err != nil {
+				return fmt.Errorf("failed to read metadata.json at %s: %v", path, err)
+			}
+
+			if err != nil {
+				return fmt.Errorf("failed to read metadata.json at %s: %v", path, err)
+			}
```
scriptnull, line 86: Let us remove the duplicate `if` block here

scriptnull (commented):
I performed a thorough review and came up with https://github.com/hasura/ndc-hub/pull/463 to express what I was looking for. Please take a look at that PR and incorporate the changes from there as you like.

PR #485, opened 2025-03-04
Release python connector with tests

SandeepSamba:


SandeepSamba:
This is a sample PR to be used as an example for connector authors to add a test config for a new connector release.

hgiasac (approved):

PR #508, opened 2025-03-15
Use unexploited 'tj-actions/change-files' action

daniel-chambers:
There has been a security exploit in the `tj-actions/changed-files` github action. Apparently it's been hijacked with code that potentially could exfiltrate secrets from CI pipelines by printing them to build logs. More info: 
* https://github.com/tj-actions/changed-files/issues/2463 
* https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised

I've changed the tag we're using to a specific git commit before the exploit code was added.

### DIFF
--- a/.github/workflows/registry-updates.yaml
+++ b/.github/workflows/registry-updates.yaml
@@ -42,7 +42,7 @@ jobs:
 
       - name: Get all connector version package changes
         id: connector-version-changed-files
-        uses: tj-actions/changed-files@v44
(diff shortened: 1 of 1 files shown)


daniel-chambers:
Looks like this has been resolved at the source, so no changes are now required as the affected tags have been fixed and the malicious code expunged: https://github.com/tj-actions/changed-files?tab=readme-ov-file#changed-files.

PR #552, opened 2025-04-21
[ent-297, ent-298] ci: Validation and pipeline changes to support NDC 0.2 connectors

SandeepSamba:
Validation changes:
- If a connector uses Version 2 packaging spec:
    - Must contain `ndcSpecGeneration` field in `connector-metadata.yaml`. 
    - Cannot have CLI plugin type as `BinaryExternal`. Can only be `BinaryInline` or `Dockerized`
- If a connector uses V1 packaging spec, it cannot have CLI plugin of type `BinaryInline`

CI changes:
- Do not push the connector archive to Gcloud bucket and make the GQL call to connector hub to include the version of the connector in it, if using packaging spec V2

## Note to connector author:
- Nothing changes if you are publishing NDC 0.1 connectors. _Optionally_ you can choose to use V2 of the packaging spec. 
- If you are publishing a connector that supports NDC 0.2, you **MUST** use Packaging spec [version 2](https://github.com/hasura/ndc-hub/blob/main/connector-metadata-types/src/connectorTypes.ts#L14-L24)

### DIFF
--- a/registry/hasura/elasticsearch/metadata.json
+++ b/registry/hasura/elasticsearch/metadata.json
@@ -7,7 +7,7 @@
     "tags": [
       "search"
     ],
-    "latest_version": "v1.5.1"
+    "latest_version": "v1.7.0"
   },
   "author": {
     "support_email": "support@hasura.io",
--- a/registry/hasura/elasticsearch/releases/v1.5.2/connector-packaging.json
+++ b/registry/hasura/elasticsearch/releases/v1.5.2/connector-packaging.json
@@ -1,0 +1,14 @@
+{
+  "version": "v1.5.2",
+  "uri": "https://github.com/hasura/ndc-elasticsearch/releases/download/v1.5.2/connector-definition.tgz",
+  "checksum": {
+    "type": "sha256",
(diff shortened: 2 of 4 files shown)


codingkarthik (commented):
The PR LGTM, if we just add some tests for the YAML Marshalling and Unmarshalling logic, it would be awesome!

scriptnull (changes requested):
I am working [on a dependency](https://linear.app/hasura/issue/CPS-1787/remove-reference-of-old-get-connector-version-api-from-connector) which will need to be released before we can merge this one. So I am marking this as "Request changes" to hold merging this PR until the dependency is released.

I will signal back once we are ready to go!

scriptnull (approved):
Deployed the dependency needed for this to prod :heavy_check_mark:

awachat:
ENT-297

awachat:
[ENT-297](https://linear.app/hasura/issue/ENT-297/add-a-ci-check-in-ndc-hub-to-not-allow-binary-inline-tupe-for-ndc-spec)

PR #556, opened 2025-04-24
Release postgres v1.0.2

codingkarthik:
This is a V2 ndc-spec version release.

### DIFF
--- a/registry-automation/cmd/ci.go
+++ b/registry-automation/cmd/ci.go
@@ -453,19 +453,22 @@ func runCI(cmd *cobra.Command, args []string) {
 		fmt.Println("Successfully updated the logos in the registry.")
 	}
 
-	if ctx.Env == "production" {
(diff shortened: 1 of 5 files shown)


codedmart (approved):

registry-automation/cmd/ci.go
```diff
@@ -453,19 +453,22 @@ func runCI(cmd *cobra.Command, args []string) {
 		fmt.Println("Successfully updated the logos in the registry.")
 	}
 
-	if ctx.Env == "production" {
-		err = registryDbMutation(ctx.RegistryGQLClient, newConnectorsToBeAdded, connectorOverviewUpdates, newConnectorVersionsToBeAdded)
+	if len(newConnectorVersionsToBeAdded) > 0 {
```
codingkarthik, line 456: @SandeepSamba Added an if condition here because ndc spec generation v0.2 connectors were being skipped which resulted in the write to the registry to fail.

PR #750, opened 2026-06-15
ddn-workspace: chown DDN config dir to runtime user (10001) for GID-0-restricted clusters

tomhasura:
## Summary

Targeted permission fix for the `ddn-native-workspace` image so `ddn auth login` can write `~/.ddn/config.yaml` on **both** GID-0-forbidding clusters and OpenShift.

The baked home `/home/hasura` is owned `root:0` (OpenShift-style `chgrp -R 0 && chmod -R g=u`), so writes rely on the pod being a member of GID 0. Two cluster postures need to work:

| Cluster type | Pod identity | How it accesses `.ddn` |
|---|---|---|
| **OpenShift restricted SCC** | Arbitrary UID (not 10001), primary GID 0 | Group-0 ownership + `g=u` perms |
| **GID-0-forbidding clusters** (e.g. Gatekeeper) | UID 10001, non-zero GID | Owner (10001) access |

The fix sets `owner=10001, group=0` with `g=u` permissions on the DDN config directory only:

```dockerfile
RUN for d in .ddn .ddn-staging; do \
  if [ -d "${HOME}/$d" ]; then \
    chown -R 10001:0 "${HOME}/$d" && chmod -R g=u "${HOME}/$d"; \
  fi; \
  done
```

- `chown -R 10001:0` — UID 10001 can write as owner; GID 0 membership (OpenShift) can write via group perms
- `chmod -R g=u` — ensures group perms mirror owner perms
- Scoped to `.ddn` / `.ddn-staging` only (environment-dependent); doesn't fail when one is absent
- Placed in the final `ubuntu:24.04` stage, after `COPY --from=workspace-builder` and before `USER 1001`

## Scope note

This fixes `ddn auth login` writes into `~/.ddn`. It does **not** address copy-home read EACCES on other `root:0` subtrees (e.g. `~/.local/share/code-server`) on GID-0-forbidding clusters — those stay `g=u` and may still partial-seed (usually fine since code-server settings regenerate).

## Test plan

- [ ] Build image with `DDN_ENVIRONMENT=production` — verify `.ddn` is `10001:0` with `g=u`
- [ ] Build image with `DDN_ENVIRONMENT=staging` — verify `.ddn-staging` is `10001:0` with `g=u`
- [ ] Run as UID 10001 (non-OpenShift) — verify `ddn auth login` succeeds
- [ ] Run as arbitrary UID with GID 0 (OpenShift) — verify `ddn auth login` succeeds

Originating PromptQL thread: https://prompt.ql.app/project/hasuraql/promptql-playground/thread/a84e47c7-68da-421e-b413-b0d69fd82e7f


ddn-workspace/Dockerfile
```diff
@@ -470,6 +470,17 @@ COPY --from=workspace-builder ${HOME} ${HOME}
...
 
+# The baked home is owned root:0 (OpenShift-style chgrp/chmod g=u above), which
+# relies on the pod being a member of GID 0. On clusters that forbid GID 0 the
+# pod runs as a non-zero UID (10001) and cannot write the DDN config dir, so
+# `ddn auth login` fails to write ~/.ddn/config.yaml. Targeted fix: hand the DDN
+# config dir to the runtime user. The dir name is environment-dependent (.ddn
+# for production, .ddn-staging otherwise), so chown whichever exists and don't
+# fail the build when one is absent.
+RUN for d in .ddn .ddn-staging; do \
+  if [ -d "${HOME}/$d" ]; then chown -R 10001:10001 "${HOME}/$d"; fi; \
```
nithindv: Wouldn't this now fail on Openshift though? Do we want to check with Sandeep once on a PQL thread to see what the best fix for this would be?

tomhasura: Modified the PR so that group is kept as 0.

Tested it locally on docker, where I am passing an arbitrary user and group=0.  Result looks good:

```
=== id ===
uid=1000680000 gid=0(root) groups=0(root)
HOME=/home/hasura
=== ls -ln /home/hasura/.ddn ===
total 12
-rw-rw-r-- 1 10001 0  167 Jun 15 16:35 config.yaml
-rw-rw-r-- 1 10001 0   29 Jun 16 13:02 last_update_check_at
drwxrwxr-x 1 10001 0 4096 Jun 16 13:01 plugins
=== write test ===
WRITE OK: /home/hasura/.ddn/config.yaml
```

Also re-tested on my EKS cluster and that looks fine as well

SandeepSamba (approved):

PR #751, opened 2026-06-17
chore(images): migrate base images to private -stable aliases

rohankmr414:
Migrates Dockerfile `FROM` base images to Hasura's private `-stable` aliases at `us-docker.pkg.dev/hasura-container-images/external-images` (preserve-source layout), so bases auto-inherit the latest vuln-patched digests.

Scoped to `FROM` lines only (preserving `AS <stage>` / `--platform` prefixes). Part of the org-wide base-image migration (Phase 1, exact-match swaps).


dliub:
Closing as out-of-scope: this PR came from an exploratory org-wide base-image `-stable` migration sweep that we've scoped down to PromptQL self-hosted images only. Reopen if needed.

